Issue #24, opened 2023-08-12
Support for full list of resolutions used to train SDXL models


MoonRide303:
It should be possible to pick in any of the resolutions used to train SDXL models, as described in Appendix I of [SDXL paper](https://arxiv.org/abs/2307.01952):

```
Height Width Aspect Ratio
512 2048 0.25
512 1984 0.26
512 1920 0.27
512 1856 0.28
576 1792 0.32
576 1728 0.33
576 1664 0.35
640 1600 0.4
640 1536 0.42
704 1472 0.48
704 1408 0.5
704 1344 0.52
768 1344 0.57
768 1280 0.6
832 1216 0.68
832 1152 0.72
896 1152 0.78
896 1088 0.82
960 1088 0.88
960 1024 0.94
1024 1024 1.0
1024 960 1.07
1088 960 1.13
1088 896 1.21
1152 896 1.29
1152 832 1.38
1216 832 1.46
1280 768 1.67
1344 768 1.75
1408 704 2.0
1472 704 2.09
1536 640 2.4
1600 640 2.5
1664 576 2.89
1728 576 3.0
1792 576 3.11
1856 512 3.62
1920 512 3.75
1984 512 3.88
2048 512 4.0
```

MoonRide303:
Vanilla has it's own way here, so - closing.

PR #28, opened 2023-08-12
Changed resolutions list to those used during SDXL training

MoonRide303:
Fix proposal for issue #24.


lllyasviel:
but the current one is also official https://github.com/Stability-AI/generative-models/blob/45c443b316737a4ab6e40413d7794a7f5657c19f/scripts/demo/sampling.py#L7

MoonRide303:
@lllyasviel The one you've linked is from the demo / example code - I would go with rather go with list from research paper, but it's your call.

lllyasviel:
Thanks for the contribution. Nevertheless, we are not sure if the real SDXL model uses exactly same parameters with the paper, since SAI official demo has different parameters. Usually following their codes is safer in my opinion. We need to close this, but feel free to ask reopen if we have more evidences about the aspect ratios.

PR #82, opened 2023-08-14
fix Python 3.9

tcmaps:
| union type syntax is Python 3.10+

### DIFF
--- a/modules/model_loader.py
+++ b/modules/model_loader.py
@@ -7,7 +7,7 @@ def load_file_from_url(
         *,
         model_dir: str,
         progress: bool = True,
-        file_name: str | None = None,
+        file_name: str = None,
 ) -> str:
     """Download a file from `url` into `model_dir`, using the file present if possible.
 


modules/model_loader.py
```diff
@@ -7,7 +7,7 @@ def load_file_from_url(
         *,
         model_dir: str,
         progress: bool = True,
-        file_name: str | None = None,
+        file_name: str = None,
```
bswck, line 10: This is incorrect and introduces a mypy assignment error (_Incompatible default for argument_).

bswck (commented):
Just simply postpone annotation evaluation (as in [PEP 563](https://peps.python.org/pep-0563/#enabling-the-future-behavior-in-python-3-7)) by placing `from __future__ import annotations` on top of the file to support [PEP 604](https://peps.python.org/pep-0604).

mashb1t:
Closing as stale

Issue #125, opened 2023-08-15
Fooocus will never add anything to your generated image file.


lllyasviel:
This is an announcement that Fooocus will not implement png meta (recording the prompts and parameters in your png file). The reasons are:

1. Whether or not to record/share prompts is user’s freedom. Although a small group of advanced fans are familiar with Automatic1111/ComfyUI’s features to record prompts in images, a dominant majority of real users from industry and entertainment do not know this. Writing prompts and configurations to the image will damage those majority people’s privacy, rights, and well-being.

2. The format of PNG meta has been abused by many SD software to bind users with specific tools. This is contradictory to Fooocus’s call for “exploring new mediums of thought and expanding the imaginative powers of the human species”. Fooocus is committed to eliminate the barriers between different software/workflows and encouraging users to share prompts (or random seeds) directly by writing it down in texts.

Revision 2023 Aug 16:

We appreciate community feedbacks and please take a look at the new feature ["Fooocus Private Log"](https://github.com/lllyasviel/Fooocus/discussions/160).

theelefant:
If there's no intention to embed the generation parameters within the PNGs (though I'm unclear on the reasoning you provided above), it would be beneficial to have access to the seed used for the images, especially since it's not currently visible in the UI.

lllyasviel:
seed restoring will be implemented soon hopefully

lllyasviel:
> Too bad, this is a must have feature for me that could be opt-in. Imo it makes no difference for other tools if there's metadata or not as they can't import either. I'll still recommend Fooocus to people who want something super easy to use.

Thanks for the feedback! 
We want to make sure that every time you open Fooocus, everything is absolutely private and clean.

MonsterAzi:
This makes it difficult to recreate images, figure out why one worked, or share prompts with others. And yes, you could just manually write down the prompts, loras, resolution, and seed down every single time you generate an image, but this is obviously difficult and would go against the goal of ease-of-use.

I understand the privacy concerns, but it should at least be an option one can turn on, if they want. I am very disappointed and hope this feature is reconsidered.

ke1ne:
@lllyasviel with no forcing at all, as some alternative, I would humbly propose to generate some txt with the same filename or json (history?)
thanks for your work!

Acephalia:
> @lllyasviel with no forcing at all, as some alternative, I would humbly propose to generate some txt with the same filename or json (history?)
> thanks for your work!

+1 

[...]

PR #152, opened 2023-08-16
Added support for saving prompt information in PNG metadata

MoonRide303:
Fix proposal for issue #125 (disabled by default).


lllyasviel:
https://github.com/lllyasviel/Fooocus/discussions/160

PR #154, opened 2023-08-16
Regarding the modification to save generation information in a json file

shingo1228:
This pull request is about saving the generation information in a json file with the same name as the generated image file. This is believed to address the following requests:

1. [The design policy of Fooocus,](https://github.com/lllyasviel/Fooocus/issues/125) which is to not store anything in the generated image.
2. The opinion of some users who want to check the generation information later.

### DIFF
--- a/modules/async_worker.py
+++ b/modules/async_worker.py
@@ -1,5 +1,5 @@
 import threading
-
+import json
 
 buffer = []
 outputs = []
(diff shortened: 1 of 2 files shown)


lllyasviel:
https://github.com/lllyasviel/Fooocus/discussions/160

PR #170, opened 2023-08-17
Implementation of Seed Value Restore Feature for "Fooocus"

shingo1228:
This pull request is designed to retain the Seed value of the most recently generated image, and by clicking the "Recycle seed" button, it [restore to that Seed value](https://github.com/lllyasviel/Fooocus/issues/125#issuecomment-1679460900).
ss_001
ss_002
ss_003
ss_004
ss_005

### DIFF
--- a/modules/async_worker.py
+++ b/modules/async_worker.py
@@ -89,7 +89,7 @@ def worker():
             seed += 1
             results += imgs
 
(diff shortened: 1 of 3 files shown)


lllyasviel:
fixed in 1.0.34

PR #180, opened 2023-08-17
some import errors changed.

idiotcommerce:
insted of import modules.path 
it is unable to take the location so i changed to from modules import path, html and some extensions based on thae.

check it


mashb1t:
This is not up to date and relevant anymore. Closing the PR.

PR #309, opened 2023-08-30
Add style filtering in Style tab

cufarvid:
Introduces a style filtering feature to the "Style" tab.

### DIFF
--- a/webui.py
+++ b/webui.py
@@ -75,8 +75,16 @@ with shared.gradio_root:
                 seed_random.change(random_checked, inputs=[seed_random], outputs=[image_seed])
 
             with gr.Tab(label='Style'):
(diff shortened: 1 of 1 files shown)


onlinedear:
This code seems to have a bug, which is that the style switch does not take effect immediately. It requires clicking 'Generate Image' once, and then clicking it again for the changes to take effect. However, the direction is great, and I hope the code logic can be improved.

cufarvid:
> This code seems to have a bug, which is that the style switch does not take effect immediately. It requires clicking 'Generate Image' once, and then clicking it again for the changes to take effect. However, the direction is great, and I hope the code logic can be improved.

Thank you for your feedback! Could you please provide more specific steps or examples to reproduce the problem?

PR #313, opened 2023-09-02
1.0.45

lllyasviel:
* Reworked SAG, removed unnecessary patch
* Reworked anisotropic filters for faster compute.
* Replaced with guided anisotropic filter for less distribution.


Dcas89:
Out of interest, may I ask why was the unet_forward_patched removed?

PR #357, opened 2023-09-13
try fix lora vram

lllyasviel:


verifiedsyckozis:
Fooocus works for me MRE doesn't as it is now.  Same issue as before.  Maybe I didn't merge or whatever right.  Just happy one works. :)

PR #407, opened 2023-09-17
Add custom styles using json file

TimothyAlexisVass:
Allows users to extend the styles array by adding style dicts to custom_styles.json

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ notification.mp3
 /node_modules
 /package-lock.json
 /.coverage*
+/modules/custom_styles.json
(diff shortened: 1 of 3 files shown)


TimothyAlexisVass:
Seems like an easy thing to fix...

lllyasviel:
will merge in next release

lllyasviel (approved):

lllyasviel:
hi you may consider 
add try catch to loading so that it does not crash when json file is misedited out of format
and add that file to gitignore so that it will not be cleared by auto update system

TimothyAlexisVass:
Okay I've refactored according to your request.

Issue #432, opened 2023-09-19
[BUG Announcement] 6GB VRAM without Refiner is influenced


lllyasviel:
Due to some reasons, 6GB VRAM (like 3060) has a memory leak problem and will OOM or be very slow after you genertate first image. 

We are still looking into what is the cause and will fix as soon as possible

Currently you can solve this by using a Refiner. This bug only influence users who do not use refiner.

lllyasviel:
https://github.com/lllyasviel/Fooocus/issues/421

Issue #421, opened 2023-09-18
Regarding the slow drawing problem of Fooocus2.0.62.


hzeasy:
@lllyasviel Thank you for creating this great tool Fooocus. Since it was upgraded to version 2.0.62 today, the drawing time is obviously much longer than the previous version 1.0. If this is due to function upgrade, I can understand, but I noticed that with the same parameters, only one picture is very fast, and more than 2 pictures at a time will cause the second picture to be very slow, and I noticed It turns out that the GPU memory has been exhausted during the generation of the second picture. The slow problem should be in the video memory. Can it be optimized to the previous level? Looking forward to replying, thank you!
![46635a043505a776020083571df74d3](https://github.com/lllyasviel/Fooocus/assets/136664755/7a5ada7b-a5cd-4039-8778-6b880a68422b)
![image](https://github.com/lllyasviel/Fooocus/assets/136664755/4be1d2d5-ede8-4cf8-921a-fc56746042a7)

lllyasviel:
are you happened to open two fooocus at the same time?

lllyasviel:
user mistake - closed

hzeasy:
> are you happened to open two fooocus at the same time?

@lllyasviel I'm sure I only opened one fooocus. When making the first picture, it takes up about 4.5G of GPU memory. The second picture exceeds the 6G range and reaches about 2G of virtual memory, so it becomes slow and there are 2 incomprehensible warnings. Hope these issues can be resolved, thank you for your reply.
![1695107827621](https://github.com/lllyasviel/Fooocus/assets/136664755/ccc6d684-51bc-4fc2-812e-0231bf6c3dfa)
![image](https://github.com/lllyasviel/Fooocus/assets/136664755/621f070f-9faa-466a-ba8d-56f8c8f389d5)

lllyasviel:
from the log it shows that you have modified the codes...
can you provide a gpu curve?
my development machine is also 3060 6GB with bad driver 537 and it does not have problems

hzeasy:
> from the log it shows that you have modified the codes... can you provide a gpu curve? my development machine is also 3060 6GB with bad driver 537 and it does not have problems

@lllyasviel Hello, in order to see the GPU curve more intuitively, I recorded a video. Please go to the network disk to download and view it.
Link：https://pan.baidu.com/s/1VQ_TPdKN2J5aR4QKDrgRwg?pwd=akai 
extract code：akai 

[...]

PR #437, opened 2023-09-19
fix vram problems

lllyasviel:


lllyasviel:
Early import of torch causes unexpected errors https://github.com/lllyasviel/Fooocus/issues/421 https://github.com/lllyasviel/Fooocus/issues/432

PR #446, opened 2023-09-20
allow users to choose path of models

lllyasviel:


Sniper199999:
why is this reverted?

PR #509, opened 2023-10-01
A simple fitler to avoid model that is not SDXL loaded

passworked:
Updated README.md:
1. Added instructions for users on how to skip model downloads.
2. Provided guidance on changing the default directory.

Updated path.py:
1. Made the default model path configurable through a variable.
2. Implemented a model filter based on regex to prevent loading of non-SDXL models.


mashb1t:
> Implemented a model filter based on regex to prevent loading of non-SDXL models.

Please consider that refiners can still be non-SDXL and update this PR accordingly if still relevant. Thanks!

mashb1t:
Closing as stale

PR #599, opened 2023-10-09
possibility of customization of WebUI / default Advanced mode & Styles in searchable Dropdown-Menu

hirnwunde:
For me it was enough if the advanced mode is activated by default and i can use the styles as dropdown with search function.

Maybe there are others who have more customizations in mind.

### DIFF
--- a/webui.py
+++ b/webui.py
@@ -14,6 +14,18 @@ import comfy.model_management as model_management
 
 from modules.sdxl_styles import style_keys, aspect_ratios, fooocus_expansion, default_styles, default_aspect_ratio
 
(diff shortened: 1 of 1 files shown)


kyeno:
I believe this PR is no longer valid, since as I can see:

```python
                advanced_checkbox = gr.Checkbox(label='Advanced', value=modules.config.default_advanced_checkbox, container=False, elem_classes='min_check')
```

There is a baked in config option for always advanced.

mashb1t:
There was the decision against using dropdowns for styles. Closing this PR.

PR #702, opened 2023-10-16
Merged some MRE features (generate hotkey, generate forever, notification sound)

MoonRide303:
Merged some non-intrusive Fooocus-MRE features (generate hotkey, generate forever, notification sound - ported from A1111).


MoonRide303:
@lllyasviel Please check if it's OK to be merged - set of non-intrusive features ported from A1111:
- generate hotkey (CTRL+ENTER),
- generate forever (via context menu available after pressing RMB on Generate button),
- notification sound (plays notification.ogg or notification.mp3 when generation is finished, if either of those files exists).

lllyasviel:
stop button does not stop generating forever - a bit bad
stop button should work and the "stop generating forever" should be removed to avoid confusing

lllyasviel:
if u do not know how to let python gradio to stop the forever loop at js level, I can take a look later
if u know how then u can just attempt a fix

MoonRide303:
That's how it originally worked in A1111 (separate starting and stopping single and forever generation), so I just left it this way. But you might be right it could be more intuitive if Stop simply stopped all generation, no matter how it was started. Maybe merge as it is, and then simplify stopping generation in another commit?

MoonRide303:
@lllyasviel Stopping simplified.

lllyasviel:
can we remove the menu binded to stop button
image
I do not think it is doing anything now

lllyasviel:
also i am not hearing the notification sound.
is it turned off by default?

MoonRide303:
@lllyasviel Yes, by default it's not played - I provided example notification file as `notification-example.ogg`, but it would have to copied to `notification.ogg` to be played (or mp3 file named `notification.mp3` - it will work, too). I use it for myself, but it could be too strong / distractive for others, so I didn't enable it by default.

MoonRide303:
And context menu removed from stop button - you're right it didn't make much sense.

lllyasviel:
cool, then add version by 1, and merge yourself after that.
Also, the next time you do not need to notify me if major diffusion algorithm and major UI are not changed.

PR #735, opened 2023-10-19
Updated requirements (Gradio, Pillow)

MoonRide303:


lllyasviel:
@MoonRide303 installation errors begin to appear in fresh download
image

all starts are several seconds slower

MoonRide303:
@lllyasviel It works fine when env is created with those versions from scratch, but it seems there is a problem with package update process - but I don't see why it failed. Could you include those versions in new release zip?

lllyasviel:
a safer way to manage versions is bigger milestone updates.
but probably reland those int number scales if they also work in current gradio.
also if only updating gradio wont cause problems, then you can update gradio (make sure to test fresh download

and even if we update gradio, the version should be 3.50.0

MoonRide303:
@lllyasviel Gradio v3.41 had major optimization related to handling events (SD web UI with earlier versions was sometimes starting for like minute, if you used multiple extensions - see https://github.com/gradio-app/gradio/issues/4841), and v3.41.2 is also used in SD web UI release v1.6.0 - pretty widely used, and tested. Pillow v9.4.0 is default version used by Colab (there is no need to downgrade it to v9.2.0).

Also if we updated those packages people could use single env for both vanilla and MRE (new Pillow is necessary for full JPG support in MRE), and save some disk space  (some users were asking for it).

lllyasviel:
gradio is added
Pillow 9.4 has some install problems on windows

MoonRide303:
@lllyasviel It might be something related to this pip wrapper invoked from launch.py - when I updated Pillow manually using pip it worked just fine:
```
(base) PS D:\tools\Fooocus> pip install Pillow==9.4.0
Collecting Pillow==9.4.0
  Using cached Pillow-9.4.0-cp310-cp310-win_amd64.whl (2.5 MB)
DEPRECATION: torchsde 0.2.5 has a non-standard dependency specifier numpy>=1.19.*; python_version >= "3.7". pip 23.3 will enforce this behaviour change. A possible replacement is to upgrade to a newer version of torchsde or contact the author to suggest that they release a version with a conforming dependency specifiers. Discussion can be found at https://github.com/pypa/pip/issues/12063
Installing collected packages: Pillow
  Attempting uninstall: Pillow
    Found existing installation: Pillow 9.2.0
    Uninstalling Pillow-9.2.0:
      Successfully uninstalled Pillow-9.2.0
Successfully installed Pillow-9.4.0
```

lllyasviel:
msvcp140.dll is in use when python script call pip and win32 set in-use files as read-only
when calling from cmd pip, msvcp140.dll is not in use and can be replaced

MoonRide303:
@lllyasviel Okay, clear - Pillow will have to wait for whole env update, then.

PR #743, opened 2023-10-20
Added more information about prompts in Private Log, made wildcards in negative prompt use different seed

MoonRide303:


lllyasviel:
about 0.1% users (perhaps less than 10 persons on this planet called Earth) of Fooocus actually uses wildcard feature
but this will confuse 99.9% normal users why there are two prompts in log

MoonRide303:
@lllyasviel Should I leave only post-processed prompt, then (raw version out)?

MoonRide303:
Corrected in https://github.com/lllyasviel/Fooocus/pull/744.

PR #748, opened 2023-10-20
Add nationality.txt wildcard

agarzon:
List made with chatGPT v4

Prompted:
`a list of all possible nationalities, one per line, no numerated, unique and never repeat. Just provide the list.`

Just a small and humble contribution 😋

### DIFF
--- a/wildcards/nationality
+++ b/wildcards/nationality
@@ -1,0 +1,195 @@
+Afghan
+Albanian
+Algerian
+American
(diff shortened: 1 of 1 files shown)


MoonRide303:
While I don't trust ChatGPT as reliable source of information (I caught it providing false information as facts few times) it looks good enough as initial version - to be corrected / updated in future, if needed.

PR #754, opened 2023-10-21
match histograms after inpaint/outpaint.

frankyifei:
code modified from "Apply color correction to img2img results to match original colors." from Automatic1111. basically this will match the contrast of inpaint/output outcome with original image.
Some finetuned SDXL model when used with outpaint will end up with a low saturation result, such as  

with this fix,

Also, I found some SDXL model is less likely having this problem but this fix does not seem to affect their outcomes. Such as original SDXL and crystalClearXL


lllyasviel:
about this , i am going to consider similar methods but with fewer dependency or even latent method in diffusion

frankyifei:
> about this , i am going to consider similar methods but with fewer dependency or even latent method in diffusion

Good to know you are going to implement it. with this method, sometimes there is still a noticeable seal. latent method sounds better for fixing the seal.

frankyifei:
close for now

frankyifei:
I have created another pull request to adopt the code without introducing dependency.

lllyasviel:
hi can you tell what model can reproduce these problems for test

frankyifei:
> hi can you tell what model can reproduce these problems for test

It mostly happens to 2d or 2.5d, especially large area of colour. What I have tested so far. 
Bluepencil seems have a subtle colour difference. CounterfeitXL is more obvious. HoloanimeXL is very obvious. 
for realistic model, MOHAWK_v14 has a very clear colour difference. 

I disabled fooocus v2 style for all tests,  outpaint to top and bottom. 

Bluepencil
BluePencilXL
CounterfeitXL
CounterfeitXL
HoloanimeXL
HoloanimeXL
MOHAWK
MOHAWK

frankyifei:
please don't merge this for now. I'm still testing if I can fix these models by replacing VAE.

frankyifei:
I'll close this as it is not needed. After replacing MOHAWK's vae with sd_xl_base_1.0_0.9vae's vae, the colour difference is gone. totally gone. This method also works with HoloanimeXL. So there is not much Fooocus needs to do. Sometimes it's the models' fault. This commit is not needed. 
here are two outpaint result after I fixed the two models, without my commit.

PR #764, opened 2023-10-22
add commonly used 9:16 output image size

leakeyfdu:


frankyifei:
this does not make sense as the total pixels would be more than 1024*1024

PR #776, opened 2023-10-24
Refactored handling max random seed

MoonRide303:


MoonRide303:
Assumptions:
1. It should allow using same MAX_SEED as in k-diffusion (2 ** 63 - 1, 9223372036854775807).
2. It should safely wrap to 0 when this value is crossed.

Test:
image

MoonRide303:
@lllyasviel Small change, but just to be sure - can you look at this? I've noticed seed + 1 is also being used in pipeline and patch, but it seems to be working. Alternative solution to places where we need different seed could passing initialized task_rng, and drawing next number from it - I can rework it this way.

Issue #713, opened 2023-10-17
Queue Errors -> Session Hijacking


666spawn999:
Hi, 
I'm trying to setup a Fooocus intance for me and my colleagues.
Sadly, a problem with the queuing System is preventing, that we can use it together. 

**Error:**
 - Client A is generating an image and exits it's browser session during generation (sure, this isn't wanted, but it happens)
 - Client B, who is waiting in the queue, gets the preview and results from the running generation of Client A
 - while getting the results the progress bar is really fast, because the frames are already generated and just buffered...
 - all following Tasks are lagging one result because the queue mixed up the resolve of the queue/clients
 - this persists indefinitely and the service needs to be restarted to fix the queue. 

Video of Problem with 2 browsers
![Fooocus_Queue_Hijacking2](https://github.com/lllyasviel/Fooocus/assets/33587160/9ac0b7b3-d121-4c7d-b71a-209560e61117)

**Steps to Replicate:**
 1. Open 2 Browser windows (same/different Browsers or incognito, doesn't matter)
 2. start a job in each window -> Client A generates, Client B is in queue
 3. Exit or reload Client A Windows ->  Client B fills the slot and gets the results of ClientA instead of it's own task!
 4. The queue is from now on mixed up. Each Generation shows the Results of the Task before.
 5. Server Restart is needed to fix

**some more thoughts**
- I have already looked into this a little bit, and I think the Gradio queue and WebSocket is involved in this. 
- The Websocket Session_Ids are different in the browsers. But when ClientA quits, the rest of the result "stream" is send to the other connected websocket, even if the ID is different. 
- Adding a check, that only the Websocket who started the task, should get results, could be a fix?

Altogether this is a big problem that currently makes it impossible to share Fooocus Installation with others. All Forks I have tried are having the same problem. Some even more because they send Results to different clients in sync.. 

I think, this issues are related:
https://github.com/lllyasviel/Fooocus/issues/625

A fix would be highly appreciated! 
If you have some tips on this, I can also try to fix it myself, but currently I'm a little lost at this issue.

If this is a general problem with gradio and can't be fixed by the Fooocus Team maybe someone has some idead how to fix it in gradio?

(I really love Fooocus and all the features and the simplicity of it. Thanks for all your work! 
But sadly, bugs like this prevent me from using it in production or teams :( )

adelorenzo-portainer:
yup, I am experiencing the same problem.

666spawn999:
[...]

Issue #501, opened 2023-09-28
[BUG]The returned image is not currently requested


qy8502:
When I requested to draw a "a sleeping cat", I accidentally closed the page in the middle of the drawing process, meaning that I disconnected the websocket connection.
![QQ截图20230928090756](https://github.com/lllyasviel/Fooocus/assets/2696897/8a4dd3b9-03f0-4ff1-b3b9-9994a246b9e1)

Then on the page that reopens, request to draw a new piece "a flying bird". However, the result it gave me was the "a sleeping cat" requested in the previous picture.
![QQ截图20230928091036](https://github.com/lllyasviel/Fooocus/assets/2696897/f53b5af1-e488-49ed-9040-4fa528004d93)

If I ask again for a picture of "a jumping monkey", it will give me a picture of "a flying bird". I won't be able to get the correct drawing until I restart the fooocus program.
![QQ截图20230928091911](https://github.com/lllyasviel/Fooocus/assets/2696897/43d862ee-143e-49c3-bf35-793febeef0fc)

serkanp:
Same here, is there any option to reset the process queue?

barepixels:
When someone accidentally close the page in the middle of the drawing process, Fooocus in the CMD windows never receives the instruction to stop...

Anyway

Yesterday MoonRide303 added a new feature to Fooocus-MRE.   Load Last Prompt
If you ever crash, close, or reload the page, you can restore the last prompt.  This is very useful

qy8502:
> When someone accidentally close the page in the middle of the drawing process, Fooocus in the CMD windows never receives the instruction to stop... The bug is user error, not software
@barepixels 

If I deploy fooocus to the server, I have to need operations personnel to handle the error when it occurs. 
The user's error will cause the software to no longer provide normal services, or the error may be an unintentional behavior such as network failure.
Please solve this problem

barepixels:
Am thinking this is a ComfyUI issue

MoonRide303:
@barepixels It's not ComfyUI problem - Comfy uses LiteGraph for UI part, and if you close and re-open tab it restores your workspace. Fooocus uses Gradio for UI part, and it's more like missing feature than a bug. Probably could be implemented using [Gradio State](https://www.gradio.app/docs/state).

lllyasviel:
a temp fix is landed in 2.1.751

PR #826, opened 2023-10-31
Hotfix: use render output per session instead of global

mashb1t:
fixes https://github.com/lllyasviel/Fooocus/issues/713, https://github.com/lllyasviel/Fooocus/issues/501

Use state for worker_outputs in webui and work with parameter in async_worker instead of global outputs.

Tested and working, now shows "initializing" when the other window has been reloaded/closed and WS connection was closed. 

@lllyasviel We might optimize this by setting skip_prompt_processing = true until the current render job is started.


lllyasviel:
looks like a smart idea 
will take a look soon

mashb1t:
@lllyasviel now includes the code adjustments to automatically stop previous abandoned tasks

mashb1t:
@lllyasviel is it realistic for the changes to be reviewed until end of this week?

lllyasviel:
will be handled in one next version

mashb1t:
I try to keep this MR up-to-date as this is a dealbreaker for multiuser access. Having the feature allows for separate sessions which is mandatory especially when LCM is too fast for the Gradio UI to keep up and users leave the page before the UI thread has finished.

lllyasviel:
unfortunatly this looks a bigger problem and we need to use other ways to fix it. some algorithmic vars are also global.
this can be closed i think

mashb1t:
As currently only one task is processed at a time they may not impact user experience in any / a significant way as resetting them in between jobs should be sufficient.
Which specific vars ar you referring to in this case?

lllyasviel:
added in 2.1.821

mashb1t:
The only remaining behavioural difference is now that tasks will still be processed when the tab is closed / page is reloaded / connection breaks, but this could also be an intended feature as one could queue a large number of images and expect them to be saved to storage, but then closes the tab.
In this PR previous jobs will be stopped by calling the following code, making use of the queue system itself.

generate_clicked
```
if worker.image_number > 0:
    print(f'\nStopping handler for previous task')
    shared.last_stop = 'stop_previous'
    model_management.interrupt_current_processing()
```

async_worker

```
elif shared.last_stop == 'stop_previous':
    print('Previous task stopped')
    break
```

I assume the current behaviour is intended and will close this PR.
Thank you for the implementation, much appreciate it!

mashb1t:
@lllyasviel + another issue still remained which was fixed in this PR, now in a separate MR, see https://github.com/lllyasviel/Fooocus/pull/981

PR #959, opened 2023-11-15
disable log via config

mashb1t:
also hides the "History Log" link if set to true

sometimes you don't want to save all created images OR one even didn't know they were all saved in the first place, so this config option allows disabling the logging altogether


lllyasviel:
https://github.com/lllyasviel/Fooocus/commit/3b97e49dd814a43fc75a26b2495603e38b8b2f62

mashb1t:
Great, thanks a lot :)

PR #961, opened 2023-11-16
Upgrade version of Gradio to 4.3.0

herve-ves:
Hi, I have a try to upgrade the Gradio framework to V4.3 which it brings better scalability and security to Fooocus. Meanwhile, Gradio V4 support custom components so that the hack component could be more maintainable.

Because the Image component has been refactored and is totally different to the V3.X one. I developed a custom component named [LegacyImage](https://github.com/herve-ves/gradio-legacyimage) to make the releated functionality compatible with Gradio V4.3.

I am testing and evaluating this commit, it works fine with UOV, CN and inpaint. I hope my work could help to Fooocus.


retohofstetter:
Would be interested in updates, if there are any!

csokun:
@mashb1t there is a new ImageEditor component https://www.gradio.app/docs/gradio/imageeditor would that be a quicker switch?

PR #981, opened 2023-11-18
advanced params refactoring + prevent users from skipping/stopping other users tasks in queue

mashb1t:
Relates to https://github.com/lllyasviel/Fooocus/pull/826
Fixes https://github.com/lllyasviel/Fooocus/issues/713
Fixes https://github.com/lllyasviel/Fooocus/issues/501

fixes inconsistency in behaviour of stop_button and skip_button as it was possible to skip or stop other users processes while still being in queue.

Expected:
sessions are completely separate, no interference or control by other users except yourself

Actual:
one can skip and stop tasks of others

Steps to reproduce:
1. open 2 tabs
2. click "generate" in tab 1
3. click "generate" in tab 2
4. see skip/stop button in tab 2
5. click skip/stop button in tab 2, task for tab1 skips/stops

**this is also possible using x tabs, where the last tab clicking "generate" can skip all previous queued tasks**


lllyasviel:
yes this is a reasonable problem and should fix.
nevertheless this also disable me from stopping a task rightly after click generate.
previously i can click stop rightly after generate, though a bit laggy

mashb1t:
shared.last_stop could be handeled not in shared but in the AsyncTask object to achieve true separation. Then skip_clicked and stop_clicked could set the property on AsyncTask and the worker could handle the process accordingly.

What would be your proposal?

mashb1t:
@lllyasviel skipping/stopping is now completely isolated, but a user can't skip/stop before the task is processed. I tried many things but in the end it comes down to the task not being updated while still queued via click on e.g. skip, despite being passed by reference as object.
I assume that this is not possible due to gradio queue for the generate button actrion.
It could also be possible to use an unique identifier per job and then update the task in the queue if it's not updatable via state, but i didn't go down that rabbit hole.

TL;DR: isolation works, stopping/skipping is possible only when processing started

Please merge or adjust if you think it's fine or have a different solution in mind, thanks :)

webui.py
```diff
@@ -81,6 +84,7 @@ def generate_clicked(*args):
     css=modules.html.css).queue()
 
 with shared.gradio_root:
+    currentTask = gr.State(worker.AsyncTask(args=[]))
```
mashb1t, line 87: fyi this can also be empty, just to prevent future checks for e.g. processing or last_stop

thiner:
Not sure whether it's caused by this PR. But I got below error message after merged "realtime lcm" PR and this PR to the main branch locally.

```log
Traceback (most recent call last):

   File "/app/modules/async_worker.py", line 814, in worker

     handler(task)

   File "/root/.pyenv/versions/3.11.6/lib/python3.11/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context

     return func(*args, **kwargs)

            ^^^^^^^^^^^^^^^^^^^^^

   File "/root/.pyenv/versions/3.11.6/lib/python3.11/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context

     return func(*args, **kwargs)

            ^^^^^^^^^^^^^^^^^^^^^

   File "/app/modules/async_worker.py", line 124, in handler

     prompt = args.pop()

              ^^^^^^^^^^

 IndexError: pop from empty list
```

thiner:
> @thiner this feature does not introduce changes in the args list of async_worker, but uses the args for temporary transmission of currentTask. Please make sure that when merging the LCM feature you keep the `get_task` method of this feature in place as it removes the currentTask from the args list. You might have an issue in your merge commit after solving merge conflicts. Please confirm as both features are working for me in isolation.

I just fixed this empty list issue. The root cause is not the Fooocus code, it's caused by load blancer(Ingress). I deployed two Fooocus pods in k8s. By default, requests will be dispatched two each pod by Ingress based on its load balancing rules. Fooocus generates many requests during the painting task. I didn't dive deep into Fooocus, but I guess the task initialization and task invoke are separated in different requests. These requests will be dispatched to each pod randomly as well. In this case, the task invoke request may fail due to not able to find the arguments.
To fix the problem, I enabled sticky session for the Ingress of k8s. Now it's working like a charm!

modules/async_worker.py
```diff
@@ -722,6 +725,8 @@ def callback(step, x0, x, total_steps, y):
             execution_start_time = time.perf_counter()
 
             try:
+                if async_task.last_stop is not False:
+                    fcbh.model_management.interrupt_current_processing()
```
thiner, line 729: fcbh is not working here.

mashb1t: Nice catch, thx! This occurrence must have slipped.

mashb1t:
@lllyasviel i've extracted all advanced_parameters to ctrls and added a PID based handling for class patch.py (overrides) to make it possible to separate the settings for each worker and job.
This not only could enable future parallelisation with multiple GPUs, but is also IMHO a clean way of user separation in Fooocus.
Your feedback is welcome!

mashb1t:
@lllyasviel i saw you just added https://github.com/lllyasviel/Fooocus/blob/1bcbd6501bb3e5db5674a18a8a8074034cda318b/ldm_patched/modules/args_parser.py#L115 yesterday. Please re-evaluate this PR as it solves the need for global variables for advanced_parameters and removes them altogether. Thanks!

freshp124:
As an iPad user, the browser page will often refresh when returning to the tab (low memory can’t keep it alive) which prevents me from stopping a task that was running before the refresh. 

I’d like to be able to stop the activity that I kicked off before the tab refresh. Since im the only remote user, being able to disable the queue isolation would work for me. E.g. via command line arg. 

Currently, to cancel the queue, I have to launch Remote Desktop on my iPad then close/restart fooocus. I can enter this as a feature request if that’s preferable.

mashb1t:
@freshp124 I hope you can understand that this is by design so users don't cancel themselves when being in a multi-user environment and isolation was implemented to fix previous behavior.
Of course it's unfortunate that your browser tab automatically refreshes when reopening. Feel free to make a feature request, but it most likely is not going to be worked on soon, only if you or some community members assemble and form a team to inplement a PR.

Issue #948, opened 2023-11-14
Extreme Speed is very fast,but the Gradio progress bar can't keep up with the speed of image generation


mmmz28:
![image](https://github.com/lllyasviel/Fooocus/assets/58453940/86bc2872-f06b-4b32-8564-94806f4dd4f5)
![image](https://github.com/lllyasviel/Fooocus/assets/58453940/a0bf88bc-8353-484b-9fe4-a3385fdaf98e)
I used Extreme Speed for quick sieving, so the number of images was selected to be 32. However, when I looked at the terminal and saw that the Gradio progress bar was still slow (not stuck), it took 492s for the Gradio progress bar to complete. But it actually feels like the image has been generated in more than a minute.

mmmz28:
In fact, when the progress bar generates the 18th image, I can see in the terminal and folder that all the images have been generated

mashb1t:
By disabling the intermediate results you can prevent that from happening. 
See checkbox implementation here: https://github.com/lllyasviel/Fooocus/pull/1013

PR #1013, opened 2023-11-22
add advanced parameter for disable_intermediate_results (progress_gallery)

mashb1t:
Fixes https://github.com/lllyasviel/Fooocus/issues/948

prevents gradio frontend process from clogging image output of progress_gallery in high throughput scenarios such as LCM with image number >= 4


pscheit:
Tested this commit directly the last few days: It works to not show the intermediate results as expected.

But it doesnt help to work around the UI issues (linked in the mentioning issues). UI still hangs and shows errors / times out.

Maybe there is just on button needed "Reconnect to Websocket" somewhere. Cause when the UI hangs in the wrong state you only can do a full reload and lose all your settings.

PR #1027, opened 2023-11-24
Added Spanish language

danivs10:


buffalonan (approved):
All translations are correct, will be a nice-to-have

PR #1035, opened 2023-11-25
Adding langauge Translator with ar.json example

ahmad-m-abbas:


superpoussin22:
it require some python libraries

ahmad-m-abbas:
> it require some python libraries

I have added the new package to the `requirements_versions.txt` fille. And added Readme file for this.

superpoussin22:
the requirement file is not there but it could be :
google-api-python-client == 2.108.0
google-cloud-translate == 3.12.1

if I'm not mistaken

ahmad-m-abbas:
Sorry for that, added it.
google-cloud-translate==3.0.2 is enough

PR #214, opened 2023-08-19
Adjust labels, count from 1

thinkyhead:
- Change label "`Image Number`" to "`Number of Images`"
- Rename `image_number` to `number_of_images`.
- For progress message count Sampling from 1...n.
- For progress message use suffixes `1st`, `2nd`, `3rd`, or `Nth`.

PR #1050, opened 2023-11-27
Fix ordinal numeral suffix

blox-dev:
Fixes a small english mistake relating to ordinal numeral suffixes.
Before:
before
After:
after


mashb1t:
relates to https://github.com/lllyasviel/Fooocus/pull/214/files#diff-64c538b31e5ff03a06701a56b8033cdcf31b656d50075e9d2f25f30e3ca6ed6aR365-R369, but can be merged as extracted functionality

PR #1054, opened 2023-11-28
Update args_manager.py to disable launching default browser at startup

vikrantrathore:
When launching using ```python entry_with_update.py --listen```, it automatically starts a default browser. But if need to run it in headless mode without launching default browser added a new flag "--disable-browser" which changes the value of auto_launch to False. This is done so that can use without launching the browser when starting it in headless mode on terminal.

### DIFF
--- a/args_manager.py
+++ b/args_manager.py
@@ -18,6 +18,8 @@ fcbh_cli.parser.add_argument("--enable-smart-memory", action="store_true",
 fcbh_cli.parser.add_argument("--theme", type=str, help="launches the UI with light or dark theme", default=None)
 fcbh_cli.parser.add_argument("--disable-image-log", action='store_true',
                              help="Prevent writing images and logs to hard drive.")
+fcbh_cli.parser.add_argument("--disable-browser", action="store_false", dest="auto_launch",
+                    help="Disable launching the default browser at startup.")
 
 fcbh_cli.parser.set_defaults(
     disable_cuda_malloc=True,


mashb1t:
this functionality already exists, you can simply use --disable-auto-launch

see https://github.com/lllyasviel/Fooocus/blob/cca0ca704a713ab153938e78de6787609c723cad/backend/headless/fcbh/cli_args.py#L46C28-L46C43

vikrantrathore:
> this functionality already exists, you can simply use --disable-auto-launch
> 
> see https://github.com/lllyasviel/Fooocus/blob/main/backend/headless/fcbh/cli_args.py#L46

In args_manager.py in https://github.com/lllyasviel/Fooocus/blob/cca0ca704a713ab153938e78de6787609c723cad/args_manager.py#L2C6-L2C6 ``` enable_args_parsing(False)``` disables the args parsing and then in https://github.com/lllyasviel/Fooocus/blob/cca0ca704a713ab153938e78de6787609c723cad/args_manager.py#L24 auto_launch is set to True.

So the flag ```--disable-auto-launch``` does not work despite being present in command line argument, I believe it will be same with other options of cli_args.py, after testing it multiple times decided to add another flag as I did not want to change fcbh core code which might be updated from upstream. Also I decided not to change the enable_args_parsing given I do not know why it was there, probably the developer has some reason not to offer all the options of fcbh cli as it is integrated into fooocus.

Also given additional arguments to cli were added in args_manager.py, so added this flag. Let me know if you would like to enable all the options of fcbh cli and can make those changes and resubmit the PR.

mashb1t:
fcbh doesn't exist anymore in the latest version of Foocus, please double check if still relevant.

vikrantrathore:
You are correct not necessary now after moving to ldm_patched and new commits which resolves it.

PR #1105, opened 2023-12-01
Fix: presets are overriding the config.txt

limonkufu:
Hi, 

This fixes the issue where the config.txt file is not honored when a preset is used


mashb1t:
Closing as stale.

limonkufu:
Sorry didn't get the notification about your reply. The usecase for me was when I try to provide additional config on top of the presets, it doesn't work without this change as the config file is ignored as preset values are overriding it. I thought it should be the other way around if I define them in config they should be used

PR #1122, opened 2023-12-02
chore: replace custom lcm function with math.lcm

PraveenKumarSridhar:
This was done as per the todo in the code. Also, I've noticed that the Python version in `environment.yaml` is `python=3.10`. and `math.lcm` is supported in this version of Python. 

PS it's my first contribution to open source projects.


mashb1t:
@PraveenKumarSridhar this PR is outdated, please resolve if still relevant with the latest version of Fooocus.

PR #1197, opened 2023-12-05
Fix: readme; typo.

scharissis:
Just a tiny fix.


mashb1t:
Already adjusted and fixed

PR #1248, opened 2023-12-06
Better multiple instances support and some other minor fix

alastorid:
### Better multiple instances support 
This PR allows us to launch multiple instances by simply set two environment variables:
For example, I have these in one `run_0.bat`, this will launch an instance using the first GPU and listen on 7865 port.
````
set CUDA_VISIBLE_DEVICES=0
set GRADIO_SERVER_PORT=7865
````
I have another one `run_1.bat`, this will launch an instance using the second GPU and listen on 7866 port.
````
set CUDA_VISIBLE_DEVICES=1
set GRADIO_SERVER_PORT=7866
````
`run_0.bat` and `run_1.bat` can be launched at once. Both of the instances can be used independantly.
|Snapshot|Description|
|--|--|
|image|two fooocuses|
|image|two busy GPUs|

### Some other minor fix
**max_image_number**
This PR added a configurible item, `max_image_number`, in `config.txt`.
Ranged from 1 to 1000 and the default is 32.

**Update succeeded**
This PR changes the order in which messages are displayed when an update fails. Previously, both the failure and success messages were being displayed when there was an actual failure, which could be confusing.

### DIFF
--- a/launch.py
+++ b/launch.py
@@ -10,7 +10,8 @@ sys.path += [root, backend_path]
 
 os.chdir(root)
 os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"
-# os.environ["GRADIO_SERVER_PORT"] = "7865"
+if "GRADIO_SERVER_PORT" not in os.environ:
+    os.environ["GRADIO_SERVER_PORT"] = "7865"
 
 
 import platform


alastorid:
Thank you, @mashb1t , for pointing out the `--port` option. I was aware of it, but since its default value appears to be 8188, which differs from the actual port Gradio listens to (7865), I initially assumed that `--port` might not be respected and that `GRADIO_SERVER_PORT` was being used instead.
Upon checking, I can confirm that `--port` does work as intended for me.
I will proceed to close this PR.

Issue #1263, opened 2023-12-07
AMD Windows Run Error


NakanoSanku:
**Describe the problem**
Error 
**Full Console Log**
```
(base) PS C:\Users\KateT\Downloads\Compressed\Fooocus_win64_2-1-791> .\run.bat

C:\Users\KateT\Downloads\Compressed\Fooocus_win64_2-1-791>.\python_embeded\python.exe -s Fooocus\entry_with_update.py --directml
Already up-to-date
Update succeeded.
[System ARGV] ['Fooocus\\entry_with_update.py', '--directml']
Python 3.10.9 (tags/v3.10.9:1dd9be6, Dec  6 2022, 20:01:21) [MSC v.1934 64 bit (AMD64)]
Fooocus version: 2.1.824
Running on local URL:  http://127.0.0.1:7865

To create a public link, set `share=True` in `launch()`.
Using directml with device:
Total VRAM 1024 MB, total RAM 16253 MB
Set vram state to: NORMAL_VRAM
Disabling smart memory management
Device: privateuseone
VAE dtype: torch.float32
Using sub quadratic optimization for cross attention, if you have memory or speed issues try using: --use-split-cross-attention
Refiner unloaded.
model_type EPS
adm 2816
Using split attention in VAE
Working with z of shape (1, 4, 32, 32) = 4096 dimensions.
Using split attention in VAE
extra keys {'cond_stage_model.clip_g.transformer.text_model.embeddings.position_ids', 'cond_stage_model.clip_l.text_projection', 'cond_stage_model.clip_l.logit_scale'}
Base model loaded: C:\Users\KateT\Downloads\Compressed\Fooocus_win64_2-1-791\Fooocus\models\checkpoints\juggernautXL_version6Rundiffusion.safetensors
Request to load LoRAs [['sd_xl_offset_example-lora_1.0.safetensors', 0.1], ['None', 1.0], ['None', 1.0], ['None', 1.0], ['None', 1.0]] for model [C:\Users\KateT\Downloads\Compressed\Fooocus_win64_2-1-791\Fooocus\models\checkpoints\juggernautXL_version6Rundiffusion.safetensors].
[...]

Issue #1185, opened 2023-12-04
AMD Radeon 6750xt. "RuntimeError: Device type privateuseone is not supported for torch.Generator() api."


xjbar:
Whenever I run the bat file, it will load the program locally fine with the GUI, but when I click generate, it throws an error saying "RuntimeError: Device type privateuseone is not supported for torch.Generator() API."

 I changed the bat file to include the AMD GPU information as said to do in the install directions.

I have tried running this through WSL2 Ubuntu with Conda environment as well as natively on Windows through the bat file.

**Full Console Log**
Running on local URL:  http://127.0.0.1:7865

To create a public link, set `share=True` in `launch()`.
Using directml with device:
Total VRAM 1024 MB, total RAM 32524 MB
Set vram state to: NORMAL_VRAM
Disabling smart memory management
Device: privateuseone
VAE dtype: torch.float32
Using sub quadratic optimization for cross attention, if you have memory or speed issues try using: --use-split-cross-attention
Refiner unloaded.
model_type EPS
adm 2816
Using split attention in VAE
Working with z of shape (1, 4, 32, 32) = 4096 dimensions.
Using split attention in VAE
extra keys {'cond_stage_model.clip_g.transformer.text_model.embeddings.position_ids', 'cond_stage_model.clip_l.text_projection', 'cond_stage_model.clip_l.logit_scale'}
Base model loaded: C:\Users\Jonathan\Fooocus\models\checkpoints\juggernautXL_version6Rundiffusion.safetensors
Request to load LoRAs [['sd_xl_offset_example-lora_1.0.safetensors', 0.1], ['None', 1.0], ['None', 1.0], ['None', 1.0], ['None', 1.0]] for model [C:\Users\Jonathan\Fooocus\models\checkpoints\juggernautXL_version6Rundiffusion.safetensors].
Loaded LoRA [C:\Users\Jonathan\Fooocus\models\loras\sd_xl_offset_example-lora_1.0.safetensors] for UNet [C:\Users\Jonathan\Fooocus\models\checkpoints\juggernautXL_version6Rundiffusion.safetensors] with 788 keys at weight 0.1.
Fooocus V2 Expansion: Vocab with 642 words.
Fooocus Expansion engine loaded for cpu, use_fp16 = False.
Requested to load SDXLClipModel
Requested to load GPT2LMHeadModel
Loading 2 new models
App started successful. Use the app with http://127.0.0.1:7865/ or 127.0.0.1:7865
[...]

Issue #1091, opened 2023-12-01
Cannot run on AMD GPU RX 580


swapnawebdev:
**Describe the problem**
AMD gpu rx 580 issue while running

**Full Console Log**
```
 generator = torch.Generator(device).manual_seed(int(seed))
RuntimeError: Device type privateuseone is not supported for torch.Generator() api.
 ```

![image](https://github.com/lllyasviel/Fooocus/assets/148637423/edac8bf7-661a-4752-8bf4-3ebb669b2922)

66hzN125ping:
My whole pc shuts down with rx580 4gb

axnix:
You can get rid of this error by this: https://github.com/lllyasviel/Fooocus/issues/624#issuecomment-1757327660
Maybe then you'll get another error "There is not enough GPU video memory available!" - Possible fix: https://github.com/lllyasviel/Fooocus/issues/624#issuecomment-1833571337
But as I already mentioned there, it didn't fixed it for me (win10,rx570 8gb)

mashb1t:
Duplicate of https://github.com/lllyasviel/Fooocus/issues/1078

As of https://github.com/lllyasviel/Fooocus/commit/8e62a72a63b30a3067d1a1bc3f8d226824bd9283 (latest Fooocus version 2.1.857) AMD with >= 8GB VRAM is now supported.
Please try with min. 8GB VRAM allocated.

PR #1265, opened 2023-12-07
Prevent BatchedBrownianTree crash on startup on AMD

Thobro:
This solves a common issue reported in #1263, #1185, #1091 and others. A solution is to enable `cpu_tree` within the class `BatchedBrownianTree` for AMD products (or at least `privateuseone` devices, as I'm not sure what kind of other hardware is also of type `privateuseone`).

### DIFF
--- a/backend/headless/fcbh/k_diffusion/sampling.py
+++ b/backend/headless/fcbh/k_diffusion/sampling.py
@@ -70,6 +70,8 @@ class BatchedBrownianTree:
             self.cpu_tree = kwargs.pop("cpu")
         t0, t1, self.sign = self.sort(t0, t1)
         w0 = kwargs.get('w0', torch.zeros_like(x))
+        if w0.device.type == "privateuseone":
+            self.cpu_tree = True
         if seed is None:
             seed = torch.randint(0, 2 ** 63 - 1, []).item()
         self.batched = True


stainz2004:
Ima be honest i can not find where and how i can enable the cpu_tree. Can anyone tell me in which folder/file it is in so i could maybe get fooocus to work finally. 
Big thanks

JAL-code:
Still getting error even though change is made.  Path is .\Fooocus_win64_2-1-791\Fooocus\backend\headless\fcbh\k_diffusion\sampling.py.  Line 68.

> self.cpu_tree = True

Class referenced at: .\Fooocus_win64_2-1-791\Fooocus\modules\patch.py  See Line 162.

Thobro:
> Still getting error even though change is made. Path is .\Fooocus\backend\headless\fcbh\k_diffusion\sampling.py. Line 68.
> 
> > self.cpu_tree = True
> 
> Class referenced at: .\Fooocus\modules\patch.py See Line 162.

`cpu_tree` is probably being set to `False` at line 70. It should work if you apply my fix. Thus adding:
`if w0.device.type == "privateuseone":
            self.cpu_tree = True`
just below line 72.

lllyasviel:
this should be already fixed

Issue #1312, opened 2023-12-10
Set Gradio analytics_enabled to false


rrdd:
Would be great to set analytics_enabled to false instead true in current config. Necessary for EU Region.

Should be possible as mentiond here.
https://www.gradio.app/docs/interface

Was not able to find the Interface init in the code.

Thank You

mashb1t:
as of https://github.com/gradio-app/gradio/pull/947/files there ar two options:

1. launch is in https://github.com/lllyasviel/Fooocus/blob/main/webui.py#L522, where analytics_enabled="False" can be added
2. env vars are in https://github.com/lllyasviel/Fooocus/blob/main/launch.py#L12, where GRADIO_ANALYTICS_ENABLED can be added as "False"

I'd propose to add the env var to not additionally clutter launch arguments. What do you think?

PR #1315, opened 2023-12-10
Disable Gradio analytics via argument

mashb1t:
fixes https://github.com/lllyasviel/Fooocus/issues/1312


lllyasviel:
We need to evaluate the influence before implementing this since Gradio is using this to sort the priority of their development and bug fix.

If we disable this, an implicit influence is that fooocus-related Gradio bugs may have lower priority on the Gradio side, because they will see 0 use of those UI features that fooocus is using.

Also, no data is visible to Google. It is collected by Gradio, and Google Static is a service tool used by Gradio. It looks okay in my opinion if Gradio owns Gradio’s statics since they already have documents and TOS for that.

Gradio seems aware of some EU conditions and those will be turned off automatically or something if in EU, but will confirm later.

mashb1t:
Thank you for the detailed feedback. Then i'd propose to make it an argument such as `--disable-analytics` to opt out. Would this combine both flexibility and privacy for you?

PR #1340, opened 2023-12-12
feat: insightface support

alx:
image

Add insightface face swapping in `Input Image` > `Image Prompt` > `Advanced` form group.


lllyasviel:
inswapper_128.onnx is a low-resolution model unmaintained for 4 years without any updates.

I cannot get good results from this model but please let me know if there are any example indicating the model is still good.

PR #1355, opened 2023-12-12
NSFW Lora

dave20190:


mashb1t:
@dave20190 Please do not commit / contribute safetensor model files as every user pulling the repository will also have to pull them, incl. Gradio and cloud providers.
I'd recommend to use a Preset instead, where you can also define the LoRAs to download.
Reference: https://github.com/lllyasviel/Fooocus/blob/main/presets/realistic.json

Issue #1419, opened 2023-12-15
always-gpu tensor error in inpaint mode


doer20:
[Fooocus Model Management] Moving model(s) has taken 0.50 seconds
[Parameters] Adaptive CFG = 7
[Parameters] Sharpness = 2
[Parameters] ADM Scale = 1.5 : 0.8 : 0.3
[Parameters] CFG = 3.0
[Parameters] Seed = 3427956263218991440
[Fooocus] Downloading inpainter ...
[Inpaint] Current inpaint model is /Fooocus/models/inpaint/inpaint_v26.fooocus.patch
[Parameters] Sampler = dpmpp_2m_sde_gpu - karras
[Parameters] Steps = 30 - 15
[Fooocus] Initializing ...
[Fooocus] Loading models ...
Synthetic Refiner Activated
Synthetic Refiner Activated
Request to load LoRAs [['SDXL_FILM_PHOTOGRAPHY_STYLE_BetaV0.4.safetensors', 0.25], ['None', 1.0], ['None', 1.0], ['None', 1.0], ['None', 1.0]] for model [/data/models/Stable-diffusion/xl/realisticStockPhoto_v10.safetensors].
Loaded LoRA [/Fooocus/models/loras/SDXL_FILM_PHOTOGRAPHY_STYLE_BetaV0.4.safetensors] for UNet [/data/models/Stable-diffusion/xl/realisticStockPhoto_v10.safetensors] with 788 keys at weight 0.25.
[Fooocus] Processing prompts ...
[Fooocus] Preparing Fooocus text #1 ...
[Prompt Expansion] a girl, full focus, cute, intricate, expressive, highly detailed, sharp focused, cinematic, new creative color, elegant, fashionable, beautiful, futuristic, colors spread, symmetry, light, fine detail, joyful, amazing, glowing, attractive, pretty, noble, perfect, lucid, thought, winning, complex, artistic, iconic, clear, brilliant, epic, shining
[Fooocus] Preparing Fooocus text #2 ...
[Prompt Expansion] a girl, very serious, full focus, elegant, intricate, dramatic bright colors, highly detailed, innocent, mystical, enchanted, magic, scenic background, professional, artistic, winning, best, fair, beautiful, cinematic, great composition, original, dynamic, exciting, attractive, creative, positive, lively, awesome, symmetry, perfect, pretty, rational, healthy, pure
[Fooocus] Encoding positive #1 ...
[Fooocus] Encoding positive #2 ...
[...]

PR #1420, opened 2023-12-15
Fix inpaint device problem in `--always-gpu` mode.

lllyasviel:
2.1.841


lllyasviel:
https://github.com/lllyasviel/Fooocus/issues/1419

Issue #1427, opened 2023-12-15
Large aspect ratio image cut in InPaint zoom


neckaros:
**Read Troubleshoot**

[x] I admit that I have read the [Troubleshoot](https://github.com/lllyasviel/Fooocus/blob/main/troubleshoot.md) before making this issue.

**Describe the problem**
If you have a large image in the inpaint and you zoom with "S" key it is cut

![image](https://github.com/lllyasviel/Fooocus/assets/7789770/b3860949-f7fc-4912-b2f9-56bcbc6e893b)

Step to reproduce: 

- in advanced choose aspect ratio 3:1

- generate without any prompt
- Move to InPaint tab and drag the generated image
- over the inpaint and press S key

lllyasviel:
https://github.com/lllyasviel/Fooocus/pull/1428

PR #1428, opened 2023-12-15
fix zooming on large-width images

daswer123:
Hi, glad to see that my work is present in this great interface.

I saw the issue #1427  and fixed the problem.
Also, if you agree, I can clean up `zoom.js` of unnecessary functions left over from auto1111.

image

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -94,6 +94,10 @@
   overflow:inherit !important;
 }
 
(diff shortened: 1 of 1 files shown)


lllyasviel (approved):

lllyasviel:
thanks alot!

PR #1432, opened 2023-12-15
Code cleanup and small improvements for zoom in inpaint

daswer123:
I cleaned up the code that was left from auto1111 and is no longer needed, also made a couple of authoring changes and added the ability to cancel the last action on ctr+z.

Before you accept this PR, I'd like you to see if these changes fit, because what I've done is my vision, which may not fit you, and if I've done something that doesn't fit, I have no problem doing a rollback of that thing.

### New things
I added the ability to do undo action on Ctrl+z , tested on various scenarios, no problems with it. 

https://github.com/lllyasviel/Fooocus/assets/22278673/91a74ff3-4ced-41c0-be9c-2a1cc22c3ae6

### Changes
1) Changed the zoom button from **Alt** to **Shift**. As I think it is much more convenient + no problem with firefox.
Originally it was **Shift** , but because someone could use horizontal scrolling, I had to set**Alt** , in your interface there is no horizontal scrolling due to auto-expand

2) Enabled blur_promt, this option removes the focus of the prompt when the user interacts with canvas.
Without this change it is necessary to make an additional click that would remove the focus and then work with canvas.

https://github.com/lllyasviel/Fooocus/assets/22278673/762fee95-a1e3-491e-9669-0173ba08b082

### DIFF
--- a/javascript/zoom.js
+++ b/javascript/zoom.js
@@ -1,18 +1,5 @@
 onUiLoaded(async() => {
     // Helper functions
-    // Get active tab
-
-    /**
-     * Waits for an element to be present in the DOM.
-     */
-    const waitForElement = (id) => new Promise(resolve => {
-        const checkForElement = () => {
-            const element = document.querySelector(id);
-            if (element) return resolve(element);
-            setTimeout(checkForElement, 100);
-        };
-        checkForElement();
-    });
 
     // Detect whether the element has a horizontal scroll bar
     function hasHorizontalScrollbar(element) {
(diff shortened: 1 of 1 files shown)


lllyasviel:
wow great!

lllyasviel (approved):

lllyasviel:
are those also avaliable in webui 1.7?

daswer123:
Everything except canceling the last action.

There are problems with this in webui, you don't have it in your interface, that's why I added it.

lllyasviel:
what problems are in webui for undo? key conflicts?

daswer123:
No, if you drop a picture through buttons like "send to sketch" then undo doesn't work. And this problem happens only if you send to "sketch" or "inpaint sketch" via buttons

lllyasviel:
thanks so much, canvas-zoom is great!

PR #1436, opened 2023-12-16
Allow images in log.html to be opened by clicking on them

MuAlphaOmegaEpsilon:
Sometimes when scrolling through the log.html file you just want to look at a given picture with more detail.
Being able to simply click on the image to open it makes things simpler.

I also removed a residual closing ‘p‘ tag in the process.

### DIFF
--- a/modules/private_logger.py
+++ b/modules/private_logger.py
@@ -36,7 +36,7 @@ def log(img, dic, single_line_number=3):
     div_name = only_name.replace('.', '_')
     item = f'<div id="{div_name}">\n'
     item += "<table><tr>"
-    item += f"<td><img src=\"{only_name}\" width=auto height=100% loading=lazy style=\"height:auto;max-width:512px\" onerror=\"document.getElementById('{div_name}').style.display = 'none';\"></img></p></td>"
+    item += f"<td><img src=\"{only_name}\" width=auto height=100% loading=lazy style=\"height:auto;max-width:512px\" onerror=\"document.getElementById('{div_name}').style.display = 'none';\"></img></td>"
     item += f"<td style=\"padding-left:10px;\"><p>{only_name}</p>\n"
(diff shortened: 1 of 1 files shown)


Hangover3832:
Indeed, the p tags are a bit messy. May be also the p tag in the parameter list should be replaced with a simple <br> to keep the text more compact. Furthermore, Im not sure if this makes sense either inside the img tag: width=auto height=100% and style=\"height:auto;max-width:512px\". The closing </img> is also not necessary.

lllyasviel:
am willing to change but it seems a few projects have hard coded the current log format so we only change it when the trade-off outweigh highly enough
but will review later

MuAlphaOmegaEpsilon:
> Indeed, the p tags are a bit messy. May be also the p tag in the parameter list should be replaced with a simple `<br>` to keep the text more compact. Furthermore, I'm not sure if this makes sense either inside the img tag: width=auto height=100% and style="height:auto;max-width:512px". The closing `</img>` is also not necessary.

Totally agree, will try to improve a bit on the compactness.

MuAlphaOmegaEpsilon:
> am willing to change but it seems a few projects have hard coded the current log format so we only change it when the trade-off outweigh highly enough but will review later

I have other ideas in mind to drastically improve the aesthetics and functionality of the log.html file when opened from the browser, let me know if it's worth to explore already that space to make a more meaningful breaking change.

In case you want to avoid breaking tools that depend on the current format, I can arrange a secondary improved html file, and in the meantime we can mark the previous one as deprecated and allow everybody depending on it to move to transition to the new one.

Hangover3832:
What about outputting a separate json? Other projects could stick on this instead on html in the future.

lllyasviel:
please refer to https://github.com/lllyasviel/Fooocus/pull/1482

PR #1482, opened 2023-12-18
Further enhancements to the log file html layout

peterng1618:
- Dark theme
- Click on thumbnail to open the full-size image
- Metadata presented in a table layout

image


lllyasviel:
Unfortunaly the css style head location is a bit weird and it does not have ending </html>

But the main concern is that if already have a previous log, the new log has no css and in mess.

But this should be a same problem cause by not having css on top

peterng1618:
Thanks for reviewing this.

Yeah, I'm aware of the existing log issue. But it should be a one-time (day) issue, seems worth it to me. We could add a note to instruct the user to rename the old log file if it exists. Or add a check to continue using the old log format until a new log file is generated (but seems like not worth it to pollute the codebase for a one-time use?)

In any case, this is my 1st time contributing. Thank you for considering it!

lllyasviel (approved):

PR #1484, opened 2023-12-18
New run script: selectable preset at launch

peterng1618:
Added a new run script + batch file:
- Scan & list all available presets, allowing users to select a preset at launch, eliminating the need to make a new batch file for every new preset.
- Error handling for invalid inputs.
- Handling when no `presets` folder/files are found: launch without preset.

I recognize there are probably more elegant implementations to achieve this goal. This implementation is intended to be non-disruptive to the users who prefer the existing run files. Feel free to refactor as you see fit.

image

### DIFF
--- a/run_preset.bat
+++ b/run_preset.bat
@@ -1,0 +1,3 @@
+
+..\python_embeded\python.exe -s run_preset.py
+pause
--- a/run_preset.py
+++ b/run_preset.py
@@ -1,0 +1,63 @@
+import os
+import subprocess
+
+
+def list_presets():
+    preset_folder = 'presets'
+    if not os.path.exists(preset_folder):
+        print("No presets found. Launching with default config")
+        return None
+
+    presets = [f for f in os.listdir(preset_folder) if f.endswith('.json')]
+    return presets if presets else None
+
+
+def display_presets(presets):
+    print("Available presets:")
+    for i, preset in enumerate(presets, start=1):
+        print(f"{i}. {os.path.splitext(preset)[0]}")
+
+
+def select_preset():
+    presets = list_presets()
+
+    if presets is None:
+        return None
+
+    display_presets(presets)
+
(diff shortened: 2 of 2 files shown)


lllyasviel:
Unfortunately this was an already rejected proposal since two months ago or so

siriume:
> Unfortunately this was an already rejected proposal since two months ago or so

 In fact, what we really need is to switch presets within the webui.

mashb1t:
the config dict can easily be updated after startup by using https://github.com/lllyasviel/Fooocus/blob/7b5bced6c2e98efa5c013624f6466ed0ba76970c/modules/config.py#L84-L95, but previously used config values have to also be changed everywhere again, which results in every element using a value of the config also having to be re-rendered.
=> Currently it's easier to just restart the whole application.

mashb1t:
> > Unfortunately this was an already rejected proposal since two months ago or so
> 
> In fact, what we really need is to switch presets within the webui.

I've implemented a UI-based solution, which still keeps the preset set in args as default.
You can find it here: https://github.com/lllyasviel/Fooocus/pull/1570

siriume:
> the config dict can easily be updated after startup by using

nice!!!

MindOfMatter:
Hi 

I just created a fork fooocus project version just synced with the original project in main branch
https://github.com/MindOfMatter/Fooocus-MindOfMatter-Edition

I created a pr in my fork project to show what we need to change to enable the feature

https://github.com/MindOfMatter/Fooocus-MindOfMatter-Edition/pull/6

If you want to test it, you need to replace your fooocus local folder with mine
(according to selected feature branch)

Also my dev branch contains all merged tested pr features in my fork project
https://github.com/MindOfMatter/Fooocus-MindOfMatter-Edition/tree/dev

I hope that it could help you :) 

Have a nice day

PR #1503, opened 2023-12-19
Add Array Support on main prompt

flannerybh:
Idea to use arrays in the main prompt.  
To use arrays, put all values inside double brackets.  For example:
`[[tall, short, fat, thin]] man.`
would produce 4 images (tall man, short man, fat man, thin man).  You also need to set your number of images to map, if you set the number of images higher, the array will just wrap around.  

You can also use multiple arrays for example:
`[[beautiful, ugly]] [[man, woman]] wearing a [[red, white, blue]] suit`
would produce all combinations of each array (you'd need to set the image count to 12).

This can be used in conjunction with the "Freeze Seed" checkbox, so each image uses the same random seed so your images will look very similar except for the differences that come from the words in the array.


xhoxye:
What is the difference between it and wildcards?

flanneryhbh:
> What is the difference between it and wildcards?

A wild card is chosen at random.  There's not guarantee that you would cycle through them all, or that you wouldn't use the same value twice.  An array will deterministically cycle through all values.

It's also much quicker to use arrays because they can be added/edited right in the prompt rather than in a config file.

flanneryhbh:
@lllyasviel , just wanted to check if this was on your radar.  This is my first contribution, let me know if there's something else I should do to move it along.  Or if you're not interested in pulling.

Thanks!

CraigOndrak:
Hi Guys, can this be please reviewed/merged so we can start using this amazing feature.

flanneryhbh:
@mashb1t is there anything I can do to get this approved/merged.  Looks like a few people are interested.

andynormancx:
I've given it a test and it works as advertised, awesome. Just what I wanted.

A nice extra feature would be some sort of auto image number setting. I would be handy if there was a checkbox to mean it would automatically run the number of images to match the number of combinations.

I guess it might be good if the Generate button told you how many images it was about to create "Generate 12 images". But I'm guessing that would involve changing when the prompt is parsed, I've not really looked at the code in any detail yet.

And a clearly much more involved feature would be to have a label under the output images showing you the array combinations used for that image.

almfisch:
> Idea to use arrays in the main prompt. To use arrays, put all values inside double brackets. For example: `[[tall, short, fat, thin]] man.` would produce 4 images (tall man, short man, fat man, thin man). You also need to set your number of images to map, if you set the number of images higher, the array will just wrap around.
> 
> You can also use multiple arrays for example: `[[beautiful, ugly]] [[man, woman]] wearing a [[red, white, blue]] suit` would produce all combinations of each array (you'd need to set the image count to 12).
> 
> This can be used in conjunction with the "Freeze Seed" checkbox, so each image uses the same random seed so your images will look very similar except for the differences that come from the words in the array.

Perhaps it makes sense to use a syntax that is also used in other systems.
Then you can use prompts more easily in different systems without having to modify them too much.

In ComfyUi the syntax is like this:
{red|blue|yellow|green}

In Automatic1111 the same, but without brackets:
red|blue|yellow|green

Ev-Bro:
This doesnt seem to be working in the current version.

andynormancx:
I’ve made my own additional changes to this to make it more flexible. They allow it to be used for applying arrays to prompt weighting for example. And also to be able to have an empty value as one of the array elements.

Not sure what the proper way of contributing this would be, I’m not up to speed on the workflow of PRs on public GitHub repos.

mashb1t:
@andynormancx you can make a PR to https://github.com/flanneryhbh/Fooocus-add-array-to-propt/tree/main so your changes will be reflected in this PR if @flanneryhbh accepts them.

flanneryhbh:
> This doesnt seem to be working in the current version.

@Ev-Bro this feature is still in a Pull Request, meaning it's not in the current base version of Fooocus.  If you are having trouble getting it to work in my forked branch, please elaborate.

Ev-Bro:
> > This doesnt seem to be working in the current version.
> 
> @Ev-Bro this feature is still in a Pull Request, meaning it's not in the current base version of Fooocus. If you are having trouble getting it to work in my forked branch, please elaborate.

My bad! Please ignore.

metercai:
> > > This doesnt seem to be working in the current version.
> > 
> > 
> > @Ev-Bro this feature is still in a Pull Request, meaning it's not in the current base version of Fooocus. If you are having trouble getting it to work in my forked branch, please elaborate.
> 
> My bad! Please ignore.

This feature is merged on the other branch version and adds the ability to group words with wildcards，https://github.com/metercai/SimpleSDXL
SimpleSDXL_2024_03_04

modules/async_worker.py
```diff
@@ -376,11 +377,15 @@ def handler(async_task):
 
             progressbar(async_task, 3, 'Processing prompts ...')
             tasks = []
+            task_rng = random.Random(seed) # may bind to inpaint noise in the future
```
mashb1t, line 380: @flannerybh why has the task_rng calculation been moved outside the for loop?

flannerybh: I don't remember exactly what I was thinking there. Your did on my fork looks reasonable. I'm busy now but I'll look back and see if there was a problem another time.  What exactly was the problem with moving it outside? I was still seeing different words before when using multiple images.

mashb1t: Let's imagine the following scenario:
- image number: 2
- prompt: `[[bird, elephant]], __color__`
- seed: 1234

Screenshot 2024-02-25 at 21 56 56


this will differ from the following prompt & config:
- image number: 1
- prompt: `elephant, __color__`
- seed: 1235

Screenshot 2024-02-25 at 21 58 23

The issue is that the images differ when generating an image with the same task_rng as the wildcard changes in a different manner (both times with seed 1234) than when using a separate seed for each task.
When generating only one image the task_rng seed is set correctly (1235) and you're getting a different result.

When using `task_rng = random.Random(task_seed)` this can be prevented:

Screenshot 2024-02-25 at 22 01 38

flannerybh: I remember now, the intention was to be able to freeze the seed, but still get different values from the wildcard. I understands the inconsistency though

PR #1521, opened 2023-12-20
Update readme.md, mention Python 3.11 support, fix a typo

kyeno:
This PR mentions Python 3.11 support aside from 3.10 (tested and works perfectly on Debian-based system) as well as fixes a single typo.


mashb1t:
Still need to test Python 3.11 on Linux.

Issue #950, opened 2023-11-15
NSFW filter


Xerxemi:
**Is your feature request related to a problem? Please describe.**

I'm currently setting up an event for a school and I'd ideally like to have some extra insurance. An NSFW filter would be useful just in case anything pops up (I don't want to get sued ;).

**Describe the idea you'd like**

NSFW filter, equivalent to that in diffusers / other webuis.

mashb1t:
@Xerxemi please check out my up-to-date fork including an NSFW filter, whoch is activatable via config.
https://github.com/mashb1t/Fooocus

tiancool:
> @Xerxemi please check out my up-to-date fork including an NSFW filter, whoch is activatable via config. https://github.com/mashb1t/Fooocus

How did you do it? I didn't see the NSFW filter in your configuration file

mashb1t:
@tiancool i copied the censor code from https://github.com/AUTOMATIC1111/stable-diffusion-webui-nsfw-censor/blob/master/scripts/censor.py
and modified it, see https://github.com/mashb1t/Fooocus/blob/main/modules/censor.py

tiancool:
I hope the author can consider making it a checkbox, in the advanced settings, to enable or disable

mashb1t:
@Nebhg this is a checker, so feel free to negate the condition.
To get more NSFW images i'd propose to use a negative prompt with "sfw" and increase its weight to a suitable amount.

daniellambert95:
@Nebhg I'm a bit confused, is it possible to generate NSFW images with Fooocus? Can you please explain how I can do this if it possible?

mashb1t:
@daniellambert95 in general, you can generate everything in Fooocus and are not limited. What limits you is the capability of the model, e.g. SDXL Base doesn't have an extensive amount of nudity in its training dataset but is capable of it when fine tuning a model via Dreambooth or a LoRA/LyCORIS/Locon/Loha etc.
You can find a comparison here: 
![image](https://github.com/lllyasviel/Fooocus/assets/9307310/981e13a3-2707-4d30-93c6-e8a977f7a7e3)

daniellambert95:
@mashb1t thanks for the information. I'm a bit of a noob to all this, could you recommend a place to start learning how to implementing/fine-tuning these models?

Nebhg:
@mashb1t When training the fine tuned mode is the text embedding vital in the model understanding new words?

Nebhg:
[...]

PR #1570, opened 2023-12-23
Add preset selection to Gradio UI (session based)

mashb1t:
This PR implements a preset selection via radio buttons, which only affects the current gradio session and does not change settings globally.
To achieve this, the logic of the recently implemented meta_parser is used to update the elements in gradio. If a preset file doesn't include specific keys mentioned in modules.config.possible_preset_keys, the initial config is used (incl. preset set in args).
Also directly downloads requirements of a specific preset, allowing usage without having to restart the app (very helpful for Colab).

This offers the flexibility to change all preset keys without having to reload the UI while still keeping the preset set on application start as default/fallback.

Improves https://github.com/lllyasviel/Fooocus/pull/1484


xhoxye:
Thank you so much. I would like to ask what this means.
QQ截图20231224085023

mashb1t:
Sorry, i do not completely get what you are referring to.
Which two toggles do you mean that have no effect?

As of https://github.com/lllyasviel/Fooocus/tree/main/presets the "turbo" preset is not a default preset and you most likely have created this yourself, please double check the content there.
In general, only keys of the preset mentioned in https://github.com/lllyasviel/Fooocus/blob/main/modules/config.py#L318 will be used to import settings for preset,s so the key "Styles" from the file you mentioned will not be taken into account.

All options from https://github.com/lllyasviel/Fooocus/blob/main/modules/config.py#L318 (except model download) are mapped to their specific meta identifier counterpart, so they can be set using the same logic as the meta prompt parser, which is intended. Nevertheless, there should not be any overlaps between the features which are user facing.

Please elaborate where exactly you have issues and send the content of your turbo.json preset if this one causes issues for you.

xhoxye:
Regardless of turbo, the resolution and style of other presets cannot be loaded.

mashb1t:
thank you for the feedback, i'll double-check tomorrow

mashb1t:
@xhoxye both resolution and style are now processed correctly + i've also added style sorting after preset load.
Feel free to test again, your feedback is welcome!

xhoxye:
It already loads normally. 
There is a small problem, but it has nothing to do with your code, that is, the "default_refiner": "" in realistic.json will cause the task to cannot be started after the switch, and it needs to be modified to "default_refiner": "None" in order to start the task normally.

Amit30swgoh:
> @xhoxye both resolution and style are now processed correctly + i've also added style sorting after preset load. Feel free to test again, your feedback is welcome!

THANKS btw I did not understand How i can always run dreamshaperXL-Turbo with all the currect settings like cfg 1.4 sharpness 3 sampler dpmmeSDE etc eg by defult please?

mashb1t:
> THANKS btw I did not understand How i can always run dreamshaperXL-Turbo with all the currect settings like cfg 1.4 sharpness 3 sampler dpmmeSDE etc eg by defult please?

I think what you're looking for is a combination of this PR and https://github.com/lllyasviel/Fooocus/pull/1580, as Turbo is not included here.
For now, it might be easiest for you to apply the changes of both PRs or to download my fork https://github.com/mashb1t/Fooocus including both features already (by changing the git remote URL or downloading code directly).

Amit30swgoh:
> > THANKS btw I did not understand How i can always run dreamshaperXL-Turbo with all the currect settings like cfg 1.4 sharpness 3 sampler dpmmeSDE etc eg by defult please?
> 
> I think what you're looking for is a combination of this PR and #1580, as Turbo is not included here. For now, it might be easiest for you to apply the changes of both PRs or to download my fork https://github.com/mashb1t/Fooocus including both features already (by changing the git remote URL or downloading code directly).

Ty again closing.

mashb1t:
@Amit30swgoh this PR is not about NSFW, please add your questions/comments to https://github.com/lllyasviel/Fooocus/issues/950.

peterng1618:
It's working very well for me. Thanks so much!

Here's some feedback:
- When switching between presets, the `Image Number` should remain rather than resetting to 1.
- A new preset option, similar to the `initial`, should be added to indicate that the current configurations are parsed from the clipboard. This isn't a major issue, but it would clarify the interface.

I'd also like to suggest some new features:
- An option in the user interface to save the current configurations as a new preset.
- It would be helpful if tooltips displayed the contents of a preset when you hover over it.
- It'd be better if the user-inputted prompts (both positive and negative) were kept when switching presets. Only the preset's default prompts get appended or removed. This is just my own pref, and I understand others might think differently.

Thanks again!

mashb1t:
@peterng1618 all good ideas, but let's keep this MR simple and extend the functionality after it has been merged.

xhoxye:
Does it conflict with 2.1.861?

mashb1t:
@xhoxye no, why should it?

xhoxye:
> @xhoxye no, why should it? 
This was my mistake because I had a conflict when trying to merge into my branch, but I resolved it later.

3Diva:
How do you enable the presets? My version of Fooocus looks like this:
presets missing temp flie

mashb1t:
@3Diva this PR is still awaiting review and has not yet been merged. You can already use the feature by checking out the git branch `feature/add-preset-selection` or `main` of my fork https://github.com/mashb1t/Fooocus by adding a git remote to your local repo OR you can use it by manually applying the file changes of this PR to your local installation.

3Diva:
@mashb1t Oh thank you for the link, I appreciate it. I'm looking forward to checking out your fork of Fooocus! :D The presets sound like a great idea! I often use LCM models and Turbo models with Fooocus and it can be a pain to set up and change all the settings every time. Having presets for a quick one button enable of those settings would be amazingly and SO HELPFUL! 😊

I hope that a preset system gets added to vanilla Fooocus too, even if the creator of it chooses to hide it behind the "Advanced" tab, it would be nice to have presets available in Fooocus.

xhoxye:
2.1.864 Could not load "default_overwrite_step" this parameter?

Issue #1607, opened 2023-12-26
Bug: Describe feature can terminate image generation of other sessions


Minekonst:
**Read Troubleshoot**

[x] I admit that I have read the [Troubleshoot](https://github.com/lllyasviel/Fooocus/blob/main/troubleshoot.md) before making this issue.

**Describe the problem**
Sometimes using the "Describe this image into Prompt" feature will terminate the image generation of of other sessions. 
I can reproduce this bug the using to sessions A and B.

1. Session A starts image generation
2. Session B starts image generation and the uses the "Describe this image into Prompt" (all while A still generates the images)

This should result in session B getting the prompt and then B starting to generate while As generation is terminated because of an error. B will generate the original prompt and not the one described from an image, as expected.
Although the behavior in the steps to reproduce are not what a typical user might do since I encountered the error also in more realistic situations.
While this bug can happen when a single user uses multiple sessions, doing the above steps this is also a reliable way to "steal" the first queue position from other users.

**Full Console Log**
[log.txt](https://github.com/lllyasviel/Fooocus/files/13773057/log.txt)

mashb1t:
@Minekonst thank you for reporting, can confirm the issue.

It works fine when using `--always-gpu`.

After some debugging, it doesn't help to set the offload device (or also load_device) directly in https://github.com/lllyasviel/Fooocus/blob/main/extras/interrogate.py#L38-L39 to get_torch_device(), so i assume that when blip is done and offloaded, the image model is also moved to the offload device while still being needed in VRAM.

The debug logs in https://github.com/lllyasviel/Fooocus/blob/main/extras/interrogate.py#L50 and below also confirm this behavior:
```
        print('loading BLIP')
        model_management.load_model_gpu(self.blip_model)
        print('loaded BLIP')

        gpu_image = transforms.Compose([
            transforms.ToTensor(),
            transforms.Resize((blip_image_eval_size, blip_image_eval_size), interpolation=InterpolationMode.BICUBIC),
            transforms.Normalize((0.48145466, 0.4578275, 0.40821073), (0.26862954, 0.26130258, 0.27577711))
        ])(img_rgb).unsqueeze(0).to(device=self.load_device, dtype=self.dtype)

        print('captioning with BLIP')
        caption = self.blip_model.model.generate(gpu_image, sample=True, num_beams=1, max_length=75)[0]
        print('captioned with BLIP')
```

```
[...]

PR #1608, opened 2023-12-26
Skip free_memory for describe (interrogate)

mashb1t:
Closes https://github.com/lllyasviel/Fooocus/issues/1607

This does prevent SDXL from being unloaded while still being used by users in parallel.
Another solution would be to enable the queue for describe click action.


lllyasviel:
this may break lowvram blip

mashb1t:
Thamk you for the fix 👍

PR #1616, opened 2023-12-27
make max image number configurable

mashb1t:
See https://github.com/lllyasviel/Fooocus/pull/1248
See https://github.com/lllyasviel/Fooocus/discussions/1511

Uses default_max_image_number as upper limit for default_image_number while keeping the default behaviour if not explicitely adjusted.

⚠️ Increasing default_image_number to a high value (depending on performance selection and GPU, roughly around > 5-10) may cause Gradio to lag heavily though.
Therefore best used in combination with https://github.com/lllyasviel/Fooocus/pull/1013.

### DIFF
--- a/modules/config.py
+++ b/modules/config.py
@@ -243,10 +243,15 @@ default_advanced_checkbox = get_config_item_or_set_default(
     default_value=False,
     validator=lambda x: isinstance(x, bool)
 )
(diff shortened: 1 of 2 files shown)


siriume:
I think it might be better to add it to presets at the same time, lots of them don't know how to modify it.

PR #1710, opened 2024-01-02
Colab: Balances the data distribution

Peppe289:
# Analyses

After conducting several tests on Google Colab and reviewing the documentation, initiating "fooocus" with specific flags not only enables the utilization of cloud computing VRAM (15GB of VRAM), significantly reducing the consumption of system RAM, but also enhances processing speed and prevents premature process termination.

## What the user notices

By default, VRAM remains unused, and the program processes solely with system memory (12GB). The program crashes when attempting to use more than one image in the prompt. The process was terminated with `^C`.

#### After this commit

Using the extra 15GB of vram allows you to make the most of the program on colab. The use is perfectly balanced and manages to process with 4 images in the prompt.

_Thanks for your work, best regards._


mashb1t:
Thank you for the contribution. While i totally agree with adding `--always-high-vram` we should not set `--disable-offload-from-vram` by default as switching models will cause them also to stay on VRAM, which may cause problems.
Did you test using Fooocus on Colab incl. switching models?

Peppe289:
> Thank you for the contribution. While i totally agree with adding `--always-high-vram` we should not set `--disable-offload-from-vram` by default as switching models will cause them also to stay on VRAM, which may cause problems. Did you test using Fooocus on Colab incl. switching models?

If you mean advanced->model, no. I kept the default settings. For the two flags, even though I kept them I didn't notice any problems.

mashb1t:
fyi if you do not offload from VRAM and switch models, every additional model will also be kept in VRAM, which is fine.
Just checked the code, this shouldn't be an issue as in https://github.com/lllyasviel/Fooocus/blob/0c4f20a0d261d4fe045aae6e5a0fcad185ba6804/ldm_patched/modules/model_management.py#L357-L359
models are unloaded when more VRAM is required than is free.
Please nevertheless test with multiple models and provide your results to make sure everything is working as expected. Looking forward to it :)

Peppe289:
> fyi if you do not offload from VRAM and switch models, every additional model will also be kept in VRAM, which is fine. Just checked the code, this shouldn't be an issue as in
> 
> https://github.com/lllyasviel/Fooocus/blob/0c4f20a0d261d4fe045aae6e5a0fcad185ba6804/ldm_patched/modules/model_management.py#L357-L359
> 
> models are unloaded when more VRAM is required than is free.
> Please nevertheless test with multiple models and provide your results to make sure everything is working as expected. Looking forward to it :)

Sorry for the delay. I have tested the models and everything seems to work correctly. Regarding the flags I still have to see, but even if they are both there it doesn't seem to cause any kind of problem

Peppe289:
> The solution of this issue has been referenced to and has helped countless users already in enabling them to use Colab. A final test should be conducted and a decision needs to be made if the changes will be merged to main or not.

okay, thanks for considering this change. best regards.

poor7:
@mashb1t I’ll add that the keys `--vae-in-fp16 --unet-in-fp16 --all-in-fp16` further increase generation speed by ~10-20% and reduce memory consumption by ~10-20%, improving overall performance and stability.

nightowL821:
Hello everyone, I'm new to this. It shows Connection Errored out every time I try to generate an image. Any solution?
Screenshot (138)

mashb1t:
@nightowL821 You're not using Colab, so please check out https://github.com/lllyasviel/Fooocus/blob/main/troubleshoot.md and if this doesn't work open a new discussion.

PR #1718, opened 2024-01-02
fix --disable-in-browser

cocktailpeanut:
Currently the `--disable-in-browser` flag is not working because the [set_defaults](https://github.com/cocktailpeanut/Fooocus/blob/4ffe8738afab7571b81864f0c19dc31385fcec80/args_manager.py#L23-L27) line is overriding the `--disable-in-browser` handling that took place here: https://github.com/cocktailpeanut/Fooocus/blob/4ffe8738afab7571b81864f0c19dc31385fcec80/ldm_patched/modules/args_parser.py#L123-L124

To fix this, just add the same handling logic after the `set_defaults`.


mashb1t (approved):
lgtm 👍

cocktailpeanut:
meow @lllyasviel

Issue #1743, opened 2024-01-04
Request for feature : Preview of style on mouse over (hover)


simartem:
It would be so nice if we had a feature to preview the style with a sample image of that style, while we move the mouse over that style.

As an example : 

![image](https://github.com/lllyasviel/Fooocus/assets/148996864/db3ff042-1b75-4204-8db8-b17e87a4004e)

Do you think its possible..

Nears-115:
I think it would be a useful feature to have, but I'm curious as to how they plan to implement it.
I think it's unique if you come up with a complete idea.

I think the substance of each style is a combination of a style name, text like a prompt, and text like a negative prompt.
Personally, I like that users can create their own original styles with just a text editor and minimal knowledge.
Can you change the 3 text combinations to 4 and add image file names or URL links?
If we were to compromise the ability for users to create their own original styles, it would probably be easier to have a feature that displays prepared images according to each predefined style.
Personally, I would like to avoid such an implementation.

I don't think it would be a good idea to just add the horizontal spear and then leave it, so I'll also write a personal request related to style.

Is it possible to have a feature like image prompts that allow you to style specific images?
It's like processing the original image file and inserting prompts, negative prompts, and style names in a specific format in the metadata part.
It would be even better if the illustrations of the original images could complement the sensory parts that cannot be explained in text, similar to image prompts.
Wouldn't it be nice if they were lined up together with the styles? Then, strangely, one style and one illustration would be connected.
Instead, there is the drawback that one processed image file is required for each style, but I think it would be nice if it could be used in conjunction with the conventional one.

mashb1t:
The easiest way would be to add a link to https://github.com/lllyasviel/Fooocus/discussions/143

docppp:
Maybe not hover images, but i think static page would be sufficient:
https://github.com/lllyasviel/Fooocus/pull/1749

simartem:
I respect all other ideas but we all know it is not the easiest way to find a style while exploring a webpage out of the gui. I understand that it is not the easiest way to put sample images over all the styles, but if it is something a developer can manage to do, I believe it is the most useful. Thank you for all the developer efforts and I respect.

mashb1t:
@simartem Gradio currently doesn't support tooltips for python, so one would have to write a custom component for Gradio 3.41.2. There is one available for javascript though: https://www.gradio.app/main/docs/js/tooltip / https://www.npmjs.com/package/@gradio/tooltip?activeTab=code
[...]

PR #1749, opened 2024-01-04
Styles samples

docppp:
This PR adds "Styles samples" site with images generates with a very basic prompt - "cat" in every available style, so user can easily compare them without a need to generate images beforehand.


mashb1t (changes requested):
All of these image files are ~ 1.2 MB (avg) => 331.2 MB to additionally check out for every installation of Fooocus, which will drastically increase the overhead.
Let's just link to https://github.com/lllyasviel/Fooocus/discussions/143 as proposed in https://github.com/lllyasviel/Fooocus/issues/1743#issuecomment-1877077821 => no additional files to be downloaded on style changes, much more dynamic as we can update the style preview images on the fly.

docppp:
I dont think that 300 MB would be such a huge issue, since Fooocus users already have to download gigabytes of data in models during first startup. And those pictures are not supposed to change, so git will leave them alone. 
Maybe if i cut the size of every picture by a half? What do you think?

poor7:
>Maybe if i cut the size of every picture by a half? What do you think?
Reduce it by at least 10 times. For most pictures, nothing will change, even if their size is reduced by 50-80 times.

docppp:
> > Maybe if i cut the size of every picture by a half? What do you think?
> 
> Reduce it by at least 10 times. For most pictures, nothing will change, even if their size is reduced by 50-80 times.

Thanks for the tip, total size is reduced to ~20 MB.

kyeno:
Btw, that's not exactly mouse over style preview. :)

mashb1t:
@kyeno mouseover preview has been implemented in https://github.com/lllyasviel/Fooocus/pull/1762
@docppp please have a look at the referenced PR and check if you can combine your changes or close one of them

kyeno:
@mashb1t BEAUTIFUL! I will test it soon.
I made a quick fork that integrates a few more PRs altogether: https://github.com/kyeno/Fooocus/tree/multimerge

Not doing a PR out of it as it'll be rather huge and non-atomic, but most of this stuff is available here anyway.

mashb1t:
@kyeno [me too](https://github.com/mashb1t/Fooocus?tab=readme-ov-file#fooocus---mashb1t-improvements), but i fear that when the PRs of this repository are starting to pile up the community will scatter in their own forks though 😐

docppp:
@mashb1t PR made by @crohrer is great, but I would leave "Style Samples" page as well, so user can check exactly how image was generated with all fooocus settings. For me, at least, it seems very useful.

mashb1t:
@docppp in the end i'm not the one to decide this, but i find both useful. If you agree you can go ahead and merge the changes of the other PR into this one. Please keep in mind that the filenames have changed to 1:1 reflect the style labels.

hswlab:
The list becomes somewhat confusing after you have tried out different styles, as the labels seem to be sorted according to their last use. Is it possible to deactivate this function or perhaps even group them, as in the following example?
image

You could keep the HTML structure and just add another label for the title.
image

**Style for displaying in 2 columns**
```
#component-99 .wrap[data-testid="checkbox-group"] {
  display: flex;
  flex-direction: column;
}

#component-99 .wrap[data-testid="checkbox-group"] label {
  display: none;
}
```

mashb1t:
@hswlab changing the appearance of the whole list is not part of this PR.
I've disabled the style sorting in my fork, please check out the code at https://github.com/mashb1t/Fooocus/blob/main/modules/style_sorter.py#L40 and below to disable the behaviour to sort the list by last used (persistent). This still keeps the position change on checkbox activation though, which i intended to remain.

PR #1764, opened 2024-01-05
Excluding the math library INSTEAD using the pow() function

Cassini-chris:
scale = 1 / (self.dim_head ** 0.5) does the same as scale = 1 / math.sqrt(math.sqrt(self.dim_head))

therefore we do not need to import math here and safe some.

Recommended: Import libraries only when you need them: This will reduce the number of times that the interpreter needs to load the library's code.

### DIFF
--- a/extras/resampler.py
+++ b/extras/resampler.py
@@ -1,6 +1,4 @@
 # modified from https://github.com/mlfoundations/open_flamingo/blob/main/open_flamingo/src/helpers.py
-import math
-
 import torch
 import torch.nn as nn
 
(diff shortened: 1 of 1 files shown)


docppp:
You sure about the math? I think it should be:
https://www.wolframalpha.com/input?i=1%2Fsqrt%28sqrt%28x%29%29+%3D+x%5E%28-1%2F4%29

GuidoBartoli:
> You sure about the math? I think it should be: https://www.wolframalpha.com/input?i=1%2Fsqrt%28sqrt%28x%29%29+%3D+x%5E%28-1%2F4%29

I agree, the correct formula is `scale = 1 / (self.dim_head ** 0.25)`

mashb1t:
@Cassini-chris please double-check, thanks!

Issue #1664, opened 2023-12-30
Queue prompts


LordMilutin:
Hello! 
I would like to know if it's possible to implement a prompt queue.
For example, I have about 20 prompts that need to generate 30 images. 
Instead of waiting for the queue to finish for each prompt one by one and retyping another one, it would be awesome if we had a queue prompt option so that we could leave as many prompts as we want and leave the PC overnight to generate them, without user input in-between. 

Please let me know if this is feasible, as I think it would tremendously improve this app. 

Thanks!

mashb1t:
Hey, this is currently not possible in the UI but only via API. 
You can find examples and further information here: https://github.com/lllyasviel/Fooocus/issues/1259 & https://github.com/lllyasviel/Fooocus/issues/1496

LordMilutin:
That's unfortunate, but thank you for the response. I'll try to deal with the API somehow, but I'm not the best when it comes to that hahaha.

docppp:
I have created simple prompt queue (I'm thinking of creating PR for this, but not in this exact form as its kinda dump right now, buy it works), you can give it a try. Remember to disable auto update on startup at it will overwrite those changes.

```
diff --git a/webui.py b/webui.py
index a5138abf..1deb772d 100644
--- a/webui.py
+++ b/webui.py
@@ -23,6 +23,27 @@ from modules.ui_gradio_extensions import reload_javascript
 from modules.auth import auth_enabled, check_auth


+QUEUE = []
+
+
+def queue_add(*args):
+    QUEUE.append(args)
+
+
+def queue_start(*args):
+    if not QUEUE:
+        yield from generate_clicked(*args)
+        return
+    for arg in QUEUE:
+        yield from generate_clicked(*arg)
+    QUEUE.clear()
+    # To use every style in single prompt:
+    # for style in legal_style_names:
+    #     argss = list(args)
+    #     argss[2] = [style]
+    #     yield from generate_clicked(*argss)
+
+
 def generate_clicked(*args):
     import ldm_patched.modules.model_management as model_management

@@ -110,8 +131,8 @@ with shared.gradio_root:
                         shared.gradio_root.load(lambda: default_prompt, outputs=prompt)

                 with gr.Column(scale=3, min_width=0):
-                    generate_button = gr.Button(label="Generate", value="Generate", elem_classes='type_row', elem_id='generate_button', visible=True)
[...]

PR #1773, opened 2024-01-06
Propmts Queue

docppp:
Basic prompts queue that remembers all selected settings. If queue is greater than 0. button Generate will run multiple times with those options simulating" settings&prompting-generating" cycle made by hand.

### DIFF
--- a/webui.py
+++ b/webui.py
@@ -23,13 +23,38 @@ from modules.ui_gradio_extensions import reload_javascript
 from modules.auth import auth_enabled, check_auth
 
 
+QUEUE_PROMPT = []
+FINISHED_IMG = []
+
+
+def queue_prompt_add(*args):
+    QUEUE_PROMPT.append(args)
+
+
+def queue_prompt_start(*args):
+    FINISHED_IMG.clear()
+    if not QUEUE_PROMPT:
+        yield from generate_clicked(*args)
+    else:
+        while True:
+            try:
+                yield from generate_clicked(*QUEUE_PROMPT.pop(0))
+            except IndexError:
+                break
+    yield gr.update(visible=False), \
+        gr.update(visible=False), \
+        gr.update(visible=False), \
+        gr.update(visible=True, value=FINISHED_IMG), \
+        gr.update(value=f"Queue ({len(QUEUE_PROMPT)})")
+
+
 def generate_clicked(*args):
     import ldm_patched.modules.model_management as model_management
 
     with model_management.interrupt_processing_mutex:
         model_management.interrupt_processing = False
 
-    # outputs=[progress_html, progress_window, progress_gallery, gallery]
+    # outputs=[progress_html, progress_window, progress_gallery, gallery, queue_button]
 
     execution_start_time = time.perf_counter()
     task = worker.AsyncTask(args=list(args))
@@ -37,8 +62,9 @@ def generate_clicked(*args):
 
     yield gr.update(visible=True, value=modules.html.make_progress_html(1, 'Waiting for task to start ...')), \
         gr.update(visible=True, value=None), \
-        gr.update(visible=False, value=None), \
-        gr.update(visible=False)
+        gr.update(visible=bool(FINISHED_IMG), value=FINISHED_IMG), \
+        gr.update(visible=False), \
+        gr.update()
 
     worker.async_tasks.append(task)
 
@@ -50,7 +76,7 @@ def generate_clicked(*args):
 
                 # help bad internet connection by skipping duplicated preview
                 if len(task.yields) > 0:  # if we have the next item
-                    if task.yields[0][0] == 'preview':   # if the next item is also a preview
+                    if task.yields[0][0] == 'preview':  # if the next item is also a preview
                         # print('Skipped one preview for better internet connection.')
                         continue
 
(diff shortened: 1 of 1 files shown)


mashb1t:
discussion: see https://github.com/lllyasviel/Fooocus/issues/1664#issuecomment-1879699479 and below

mashb1t (changes requested):
see https://github.com/lllyasviel/Fooocus/issues/1664#issuecomment-1879763767 to make the code compatible for simultaneous use of multiple users. I can provide code optimisations tomorrow.

LordMilutin:
I have tried setting it, and running it, it works fine on the frontend, I am able to add prompts to queue, however, the generation does not work:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/gradio/routes.py", line 488, in run_predict
    output = await app.get_blocks().process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1431, in process_api
    result = await self.call_function(
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1103, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "/usr/local/lib/python3.10/dist-packages/anyio/to_thread.py", line 56, in run_sync
    return await get_async_backend().run_sync_in_worker_thread(
  File "/usr/local/lib/python3.10/dist-packages/anyio/_backends/_asyncio.py", line 2134, in run_sync_in_worker_thread
    return await future
  File "/usr/local/lib/python3.10/dist-packages/anyio/_backends/_asyncio.py", line 851, in run
    result = context.run(func, *args)
  File "/usr/local/lib/python3.10/dist-packages/gradio/utils.py", line 707, in wrapper
    response = f(*args, **kwargs)
  File "/stable-diffusion/modules/advanced_parameters.py", line 21, in set_all_advanced_parameters
    disable_preview, adm_scaler_positive, adm_scaler_negative, adm_scaler_end, adaptive_cfg, sampler_name, \
ValueError: too many values to unpack (expected 32)
```

mashb1t:
> I have tried setting it, and running it, it works fine on the frontend, I am able to add prompts to queue, however, the generation does not work

@LordMilutin works for me. Please check that when resolving merge conflicts the appropriate parameter count is returned in modules/advanced_parameters.py (adjust counter accordingly).

@docppp Here are even more reasons why this queue needs to be well thought out:
* advanced parameters are currently shared between generations, so when queueing a render and changing an advanced parameter in the Developer Debug Mode tab, this overrides the previous queued entries. A solution would be to set all advanced params to your queue, then call `advanced_parameters.set_all_advanced_parameters` with them on queue execution in your while loop. This does not work for multi-user scenarios though as `advanced_parameter` uses globals, so no separation whatsoever.
* Your current implementation breaks image output for when a queue has only items with image amount = 1 queued, as there is no final gallery output due to `do_not_show_finished_images=len(tasks) == 1` in async_worker.py
* Speed of the gallery is heavily impacted for large queue / image amount size. Solved by https://github.com/lllyasviel/Fooocus/pull/1013

I'd propose to mark this PR as draft, as it is not ready to be merged and might need to undergo an in-depth feasibility analysis.

LordMilutin:
I have tested it, and it works perfectly for my issue. Great job!

docppp:
good points, I will try to analyze the code more and see if I can come up with any feasible solution. but I will focus solely on single-user scenario this is mine field of interest.

mashb1t:
Thanks!
Just to clarify: Gradio generally offers all that one would need for parallel usage and multiple users etc., but Fooocus is not built for maximum speed and parallel processing, but accessibility and user-experience. When active development continues in mid/end January (see https://github.com/lllyasviel/Fooocus/blob/main/update_log.md first line) there might be new information and optimisations to integrate this feature properly.

LordMilutin:
@docppp one thing to consider, I tried queuing last night 50 queues with 32 images each, and after the 2nd or 3rd queue it stops generating. After refreshing the page and queuing again it works for a few queues and then stops. 
Is it possible to send all queues to the backend at once and process them that way even if the frontend freezes?

docppp:
@LordMilutin please check it out, with new fix, I am able to generate over 250 images and it keeps going.

blablablazhik:
@docppp hello! Sry for bothering you again, but can you add txt prompt reader in queue? So u can add 1 txt file with prompts on each line and it will generate with all settings each prompt per line

LordMilutin:
> @LordMilutin please check it out, with new fix, I am able to generate over 250 images and it keeps going.

Thanks, I was finally able to generate the whole queue last night with the update you made. Thank you so much for this!

Two things to consider:
1. I believe there should be a clear queue button as I accidentally added the same prompt to queue a few times and I had to restart the app or let it generate the same prompt a few times.
2. This version shows only the latest prompt-generated images in the UI. This isn't a huge dealbreaker as I've found previous prompts in tmp/gradio folder so I can sort it by timestamp and roughly see what prompts they belong to. I prefer this rather than the previous version that showed all prompt-generated images but the queue never completed.

webui.py
```diff
@@ -87,7 +87,9 @@ def generate_clicked(*args):
                     gr.update(visible=False), \
                     gr.update()
             if flag == 'results':
+                global FINISHED_IMG
                 FINISHED_IMG.append(product[-1])
+                FINISHED_IMG = FINISHED_IMG[-32:]
```
mashb1t, line 92: just a note, the max image amount per generation is configurable in the config and only defaults to 32, so this might be changed to the config rather the hardcoded number

docppp: I would rather leave it as is, as more items in gallery may break things. It is easy to check what was generated in history log file where prompts are also shown.
@LordMilutin yup, clear button is a good idea and about point 2, see above.

@blablablazhik prompts from file is not in my plans, but you can try hardcoding it yourself (i believe i showed somewhere how to do it).

LordMilutin: > I would rather leave it as is, as more items in gallery may break things. It is easy to check what was generated in history log file where prompts are also shown. @LordMilutin yup, clear button is a good idea and about point 2, see above.
> 
> @blablablazhik prompts from file is not in my plans, but you can try hardcoding it yourself (i believe i showed somewhere how to do it).

Upon further usage, I agree that a clear queue button is a must, as well as a stop queue button. Because right now if I stop it, it will move on to another queue, and so on, so I have to stop each queue which is a lot if I have 50 queues.

VictorZakharov:
Even single user would be great to start. Useful for experimenting with styles with batch gen, for example, to avoid waiting. Queue 100-1000 gens then come back in an hour or smth like that.

GalacticCurtain:
The final gallery isn't showing once the queue is complete, any hints how to fix that ?

Issue #1618, opened 2023-12-27
Model list in alphabetical order


GuidoBartoli:
**Is your feature request related to a problem? Please describe.**
When I scroll through the list of available models, it takes quite some time to find a specific one because the list is not in alphabetical order.

![Screenshot from 2023-12-27 22-07-29](https://github.com/lllyasviel/Fooocus/assets/16103676/be548366-364c-41c5-a323-379f4d940cfe)

**Describe the idea you'd like**
Sort the list in alphabetical order.

mashb1t:
Implemented in PR https://github.com/lllyasviel/Fooocus/pull/1517

MeterLong:
> [在 PR #1517](https://github.com/lllyasviel/Fooocus/pull/1517)中实施[](https://github.com/lllyasviel/Fooocus/pull/1517)

是不是考虑同时添加模型搜索功能呢？

mashb1t:
You can already filter the list by just starting to type. Gradio will then only show you models including the entered value in their name.

GuidoBartoli:
> Implemented in PR #1517

Nice, hoping that it will soon be merged into main branch!

Thanks

mashb1t:
Solved the root cause of the issue, fixed in https://github.com/lllyasviel/Fooocus/pull/1784

PR #1517, opened 2023-12-20
Sort Checkpoints and LoRAs alphabetically

kyeno:
This simple PR fixes the right menu's "Model" tab to ensure the Checkpoints and Loras are always sorted alphabetically.
Both at startup and after clicking "Refresh All Files" button.

PR #1784, opened 2024-01-07
fix: correctly sort files, display deepest dir level first

mashb1t:
There is an issue in the file sorting, which has to do with the condition of the lambda. `os.sep` returns correctly, but the sorting of the else condition just appends to the end and beginning, which is wrong.
This PR implements sorting for files when walking as well as `topdown=False` to display subfolders first.

Closes https://github.com/lllyasviel/Fooocus/issues/1618
Relates to https://github.com/lllyasviel/Fooocus/pull/1517

before:
Screenshot 2024-01-07 at 14 17 19

after:
Screenshot 2024-01-07 at 14 16 15

### DIFF
--- a/modules/util.py
+++ b/modules/util.py
@@ -164,14 +164,14 @@ def get_files_from_folder(folder_path, exensions=None, name_filter=None):
 
     filenames = []
 
-    for root, dirs, files in os.walk(folder_path):
+    for root, dirs, files in os.walk(folder_path, topdown=False):
         relative_path = os.path.relpath(root, folder_path)
         if relative_path == ".":
             relative_path = ""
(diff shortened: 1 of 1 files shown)


wmlele:
Sort of related to this: is there any reason (security? isolation?) for not following symlinks to directories (`followlinks:True`) in the `os.walk() `call? 

I don't think I am the only one to have an organized model store with lots of categories and symlinks to them. If nothing prevents it, could it please be enabled?

mashb1t:
@wmlele path_checkpoints and path_loras can now be an array in the config, basically there is no need for symlinks now. Please check if this works for you.

PR #1843, opened 2024-01-09
Configurable autoupdate

docppp:
`"enable_autoupdate": false` can be added to config.txt to disable auto update on run.bat.

### DIFF
--- a/entry_with_update.py
+++ b/entry_with_update.py
@@ -6,41 +6,46 @@ root = os.path.dirname(os.path.abspath(__file__))
 sys.path.append(root)
 os.chdir(root)
 
+from modules.config import enable_autoupdate
 
-try:
-    import pygit2
-    pygit2.option(pygit2.GIT_OPT_SET_OWNER_VALIDATION, 0)
+if enable_autoupdate:
+    try:
+        import pygit2
+        pygit2.option(pygit2.GIT_OPT_SET_OWNER_VALIDATION, 0)
 
-    repo = pygit2.Repository(os.path.abspath(os.path.dirname(__file__)))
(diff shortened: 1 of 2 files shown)


mashb1t:
In which scenarios is this benificial over using launch.py directly as an entrypoint?

docppp:
@mashb1t this PR is not about benefits, but freedom of choice. i believe more than 90% of users have no knowledge about python (and just use run.bat), so dont assume they would know how to properly setup venv, or how to use existing python interpreter, to run launch.py. with this modification, they can easily choose not to update fooocus.

mashb1t:
@docppp i totally get your point but using launch.py is as easy as modifying the one line in run.bat and replace `entry_with_update.py` with `launch.py`. No need to manually tinker with python or zhe config at all.
Would you see this option or the config adjustment easier/more accessible?

docppp:
Editing bat file may be overwhelming for some, even if this is very simple. Putting that option inside config,txt that is supposed to be modified is a sign that this is valid welcome modification.

mashb1t:
They are also supposed to edit the run.bat file to add & adjust argument, which is why we have them described in the [readme](https://github.com/lllyasviel/Fooocus?tab=readme-ov-file#all-cmd-flags). Using the config would also allow to enable/disable auto-update for each preset, as the keys will just override the config, which might not be intended.
This is why i personally agree with https://github.com/lllyasviel/Fooocus/pull/1843#issuecomment-1884032150 and would like to streamline the approach for using launch.py as replacement, not modifying the config, and rather add a description to the readme how to do so.
In the end, i'm not hte one merging the MR, so this decision is open to the Fooocus core dev team.

mashb1t:
Closing this PR as already pointed out. launch.py can be used as entrypoint instead.

PR #1640, opened 2023-12-29
Fooocus wildcards enhance 通配符增强，ComfyUI Portrait Master 

xhoxye:
QQ截图20240109213514

QQ截图20240109213435

QQ截图20240109213457

PR #1861, opened 2024-01-11
Wildcard Select

midareashi:
Implemented a new Wildcard tab that lists .txt file names from the Wildcard folder. Users can select wildcards from the list, automatically placing them at the beginning of positive prompts. This functionality enhances the ease of recalling wildcard file names and simplifies the process of adding them to prompts.
image
image


JulienZLQ:
Wildcard is not only used at the beginning.
Can be freely combined with other texts, which is more in line with the way Wildcard is used.

midareashi:
The issue is there is no wildcard manager. You can download hundreds of wildcard files and there is no easy way to reference this list. This makes it easier to find and sort wildcards within the Fooocus UI. You can still manually add __wildcard__ to the prompt to control where it shows up. A possible addition to this code would be adding a <copy> button to the wildcard name so you can paste it in the prompt where you want.

docppp:
I would prefer tab with wildcards preview, rather than selectable options. With this, you can easily check what wildcards are available and what hides behind then (lets say you could show first 2 or 3 elements from the set as they may be arbitrary long). Copy to clipboard on clicking seems nice.

midareashi:
I agree. I'm updating this to be a sort/filter tool with copy to clipboard onclick. That would work much better for typical wildcard use cases. 

My current code is really a prompt randomizer. I'm going to keep this but call it wildprompt. It works well if you have a collection of complete prompts you want to randomize. Wildprompt will be decouple it from wildcards - separate folder, separate tab, different functionality. I've had a lot of fun writing different prompts and randomizing them in large batches using different models and styles.

This is an example of a wildprompt file.
[1girl-fantasy.txt](https://github.com/lllyasviel/Fooocus/files/13908699/1girl-fantasy.txt)

JulienZLQ:
> I understand what you mean. I have been using wildcards to generate entire prompts, not just modify parts of prompts.
> 
> I'll update this code to copy the `__wildcard__` to the clipboard so it can be added anywhere in the prompt. I still think it's useful to sort and filter names of wildcards. Some packs contain over 1000 wildcards so management can be a hassle.

So in fact, you are using wildcards as prompt defaults. This operation can be understood, but the current function is only for such use, and cannot cover other usage methods such as regular individual content. If it can If we take both into account, I think I would also like this function very much.

mashb1t:
@midareashi  This is already implemented in https://github.com/lllyasviel/Fooocus/pull/1640 with even more functionality. Please check the changes there and feel free to suggest your additional changes (if any) there.
Screenshot 2024-01-12 at 17 56 41

midareashi:
@mashb1t That's cool, I'll check it out. If wildcards are already handled I'll just work on my prompt randomizer which was what I really wanted in the first place.

Issue #1789, opened 2024-01-07
Jpeg 100% quality save


RodTDai:
Hey guys!
Should be nice to have an option to save image in JPEG 100% quality instead of png. 
Thanks.

mashb1t:
Saving could be achieved relatively easy by changing the two lines https://github.com/lllyasviel/Fooocus/blob/176faf6f347b90866afe676fc9fb2c2d74587d7b/modules/private_logger.py#L16 and https://github.com/lllyasviel/Fooocus/blob/176faf6f347b90866afe676fc9fb2c2d74587d7b/modules/private_logger.py#L25 accordingly.
Temp images in the browser are png base64 encoded images directly from pytorch_to_numpy as far as i could see and there is no conversion happening.
If you need the jpg/jpeg results only in your outputs folder, feel free to change private_logger.py accordingly.

Enhancement: make this configurable.

RodTDai:
nice. Yes configurability is important here, because when system update files, we loose that custom modifications.
Thanks

eddyizm:
A config something like this ?
![image](https://github.com/lllyasviel/Fooocus/assets/12236892/2ade2380-fbd4-4c28-9f5f-184262c523d2)

I got the front end mostly working but need to figure out how to update the configuration using gradio if this works.

eddyizm:
tested my pr locally but did not see any unit tests. If this needs to save to the file to maintain state, I can update my PR.

drdancm:
This would be a wonderful enhancement.

RodTDai:
actually a change for gradio is a good thing, as it can become super slow when running machine at home and accessing via mobile. Jpeg will help that too.

Irmagon:
Maybe add an additional parameter _image_quality_ specifying the JPG compression ratio?
And setting this in UI input box or config.txt.
For example: `Image.fromarray(img).save(local_temp_filename, quality=image_quality)`
In PIL.Images _quality_ only affects the lossy compression format and not be used for PNG.

mashb1t:
@Irmagon already noticed when implementing https://github.com/lllyasviel/Fooocus/pull/1940, fixed in https://github.com/lllyasviel/Fooocus/pull/1863/commits/f1261028b8b4d0d57e5205cdea9755abdb404e95

Issue #1421, opened 2023-12-15
Image save filename consistency


flashthepublic:
**Is your feature request related to a problem? Please describe.**
When you save an image sometime the filename used by the browser is just "image.png" which leads to issue when you save multiple images
Sometimes its a more usable "image - _timestamp_" (ex: "image - 2023-12-15T093912.142")

**Describe the idea you'd like**
Could it be consistently the name with the timestamp? it's easier to save when you organize multiple generation in different folders

mashb1t:
Fooocus should always use generate_temp_filename from https://github.com/lllyasviel/Fooocus/blob/main/modules/util.py#L141 (tested locally and in Colab).
Please provide some examples and steps to reproduce when the filename is image.png.

mJastrzebski6:
(On Windows) Using this button the photo will be saved in the downloads folder with the name "image".
<img width="618" alt="Screenshot_1" src="https://github.com/lllyasviel/Fooocus/assets/70579449/66e36fe8-219c-4b4d-b4a2-5e13ebfa3623">

mashb1t:
@mJastrzebski6 just to make sure, are you certain you're using the latest version of Fooocus (2.1.857)?

flashthepublic:
> Fooocus should always use generate_temp_filename from https://github.com/lllyasviel/Fooocus/blob/main/modules/util.py#L141 (tested locally and in Colab). Please provide some examples and steps to reproduce when the filename is image.png.

I opened the webpage clicked generate
then clicked on one of the image and then on the icon to download.

![image](https://github.com/lllyasviel/Fooocus/assets/68439171/dae3cee6-e0fa-48e1-a43c-917ae09a8ea2)

mashb1t:
> @mJastrzebski6 just to make sure, are you certain you're using the latest version of Fooocus (2.1.857)?

@flashthepublic can you please also check?

flashthepublic:
> > @mJastrzebski6 just to make sure, are you certain you're using the latest version of Fooocus (2.1.857)?
> 
> @flashthepublic can you please also check?

Tried on 857
![image](https://github.com/lllyasviel/Fooocus/assets/68439171/8744f1f5-2349-407c-8e12-0478e5e5e2f4)
and 859
![image](https://github.com/lllyasviel/Fooocus/assets/68439171/a72d89b2-8625-41d2-a7d9-9aeb42900f06)


For information in the History log filename seems correct
[...]

PR #1863, opened 2024-01-11
Jpg config image extension

eddyizm:
[Added]
feature: added flag, config and ui update for image extension change)[feature: added flag, config and ui update for image extension toggle.

https://github.com/lllyasviel/Fooocus/issues/1789

image

image
image


mashb1t (changes requested):
Thank you for the implementation.
The config should only be read on app start and not be adjusted during runtime, as this is a global setting.
For UI elements Fooocus currently uses the AsyncJob object, which contains an array of parameters and which is introducing isolation per user/session. Using your approach (config) leads to User 1 updating the output extension also for User 2, who might not be aware / is not made aware of in the UI that the change happened and this way the feature isn't usable in isolation.

This is why i'd propose to add the UI element to the ctrls in webui.py, add a pop() call in async_worker.py and hand it over to the private_logger.py's log() method as a parameter. The radio buttons should also only be shown/interwctive when history log is enabled (see args).
The config is useful and can still be used as default for the instance.

What do you think?

eddyizm:
Hi thanks for the feedback and clarification. Did not realize the config was global so thanks for pointing that out. 
Your suggestions seem reasonable and will give it a go this evening. Appreciate the guidance. New to gradio so this seemed like a good entry point to be able to contribute.

RodTDai:
This is awsome guys, thanks for the work you are all doing.

eddyizm:
Updated with the changes requested. Please let me know if I missed something.

mashb1t (approved):
@eddyizm i've fixed a bug (missing extension param), added webp support and adjusted the label, as the label "Image Extension" could be misunderstood as images shown in the webui are still always png (see issue https://github.com/lllyasviel/Fooocus/issues/1421). The wording is still not 100% perfect, feel free to adjust to your liking.

Comparison between output formats, incl. webp (same seed & settings):
image

What do you think, should we integrate the proposed solution of https://github.com/lllyasviel/Fooocus/issues/1421 to this feature?

eddyizm:
@mashb1t I agree the naming is not correct. by using image extension we are implying format which it looks like originally I was just changing the naming convention, so more like image naming extension or file extension. When I was changing it I was looking at the gradio docs because there should really be some type of mouse over with a not or a star to say "this does not change the actual creation or compression logic, just simply changes the file extension for convenience" or something to that effect. 

Although after seeing the outputs, it does appear to be affecting the size of the file. 

Good catch on the log parameter, seems like I had it in there at some point and backed it out. 

Everything else looks good to me!

mashb1t:
@eddyizm functionality from https://github.com/lllyasviel/Fooocus/pull/1932 has been merged, works perfectly for me. Now also the preview / final images in gradio do have the selected file extension.
Maybe we can rename the label / field to "File Extension", as this plainly states what it is (both gradio and outputs).

Please check, happy to hear your feedback!

eddyizm:
Looks good to me. That last label change is perfect!

mashb1t:
@eddyizm A label and naming even better describing what it is is "Output Format". Hope this is also ok for you.

eddyizm:
Yup makes more sense. I was thinking about changing it as well. Good call.

mashb1t:
@eddyizm fyi now that https://github.com/lllyasviel/Fooocus/pull/1940 has fully been implemented i've also added EXIF metadata handling for jpg and webp, see https://github.com/mashb1t/Fooocus/commit/b281375ce26bd1f888c51e9e6b1ca91476172f45

eddyizm:
This bad boy keeps getting bigger and bigger!

mashb1t:
@eddyizm would it be ok for you if i add myself as the committing author?
13/18 commits were done my be and I've already integrated EXIF data handling for jpg and webp. 
ofc. adding you as co-author

eddyizm:
@mashb1t sure. No problem at all!

PR #1920, opened 2024-01-13
Full prompt log

docppp:
This PR adds hidden full raw prompts into the log file, so you can see applied styles in action.


mashb1t:
Is also most likely going to be in metadata handling in https://github.com/lllyasviel/Fooocus/pull/1940

mashb1t:
Hey @docppp, sorry that merging this PR took so long but nevertheless thank you for your contribution, much appreciated!
This feature has now been updated to the latest version of develop and now correctly adds the full positive and negative prompt to the log, but not to the image metadata.
Thanks again!

PR #1938, opened 2024-01-15
feat: automatically describe image on uov image upload

mashb1t:
Describes the uploaded image if prompt is empty (prevents prompt override).
This allows for better results by inexperienced users as they do not have to write a prompt from scratch when creating variations + it saves a few steps.


mashb1t:
i'm normally not keen on merging my own suggested features without review, and this one is exactly that, a suggestion. Not necessarily needed, but can be helpful.
For it to be merged i'm open to suggestions if people of the community or @lllyasviel think this improves the current situation.
Happy to hear some feedback!

aolko:
> i'm normally not keen on merging my own suggested features without review, and this one is exactly that, a suggestion. Not necessarily needed, but can be helpful. For it to be merged i'm open to suggestions if people of the community or @lllyasviel think this improves the current situation. Happy to hear some feedback!


I suggest this feature being in vanilla fooocus, as it is an essential must have

PR #1955, opened 2024-01-17
Upgrade requirements_met function

Hunter-Houts:
added more logging to see what goes wrong, helps fix random errors.

### DIFF
--- a/modules/launch_util.py
+++ b/modules/launch_util.py
@@ -5,7 +5,6 @@ import subprocess
 import sys
 import re
 import logging
-import re
 import importlib.metadata
 import packaging.version
 
--- a/modules/launch_util.py
+++ b/modules/launch_util.py
@@ -5,6 +5,9 @@ import subprocess
 import sys
 import re
 import logging
+import re
+import importlib.metadata
+import packaging.version
 
 
 logging.getLogger("torch.distributed.nn").setLevel(logging.ERROR)  # sshh...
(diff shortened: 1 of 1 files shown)


mashb1t (changes requested):
Goood changes, just a small improvement

Hunter-Houts:
> Good changes, just a small improvement

Thanks! what changes are you requesting? (or am I misunderstanding?)

modules/launch_util.py
```diff
@@ -5,6 +5,9 @@
 import sys
 import re
 import logging
+import re
```
mashb1t, line 8:
```suggestion
```

Hunter-Houts:
> I propose to replace regular expressions with [packaging.requirements.Requirement](https://packaging.pypa.io/en/stable/requirements.html)
> 
> Or... What is the purpose of `requirements_met`? Why couldn't we get rid of it entirely and call `pip install` every time unconditionally?

Thanks for the pointers! I've just pushed some changes to the PR, and here's the scoop on what I did:

1. **Switching to `packaging.requirements.Requirement`:**
   - You were spot on about using `Requirement`. It's in there now, and it's handling the version specs like a champ. It's way better for dealing with all those funky version formats in `requirements.txt`. Definitely a cleaner and more Pythonic way to go.

2. **About ditching `requirements_met`:**
   - Thought a lot about this one. I decided to keep it around for now. Mainly, it's there to keep things quick and light on network usage. It checks if we've already got the right stuff installed, so we don't waste time and bandwidth redoing `pip install`. It's kind of like, why go shopping if you’ve already got everything you need at home, right?
   - I get the appeal of just hammering `pip install` every time to keep it simple, but I’m playing the long game here for speed and efficiency.

Really appreciate your input – it kicked off some good changes. Check out the updated code and let me know what you think. Open to more ideas, always!

Hunter-Houts:
> Thanks for the changes.
> 
> I'm sorry for the second part of my comment. That was of course beyond the scope of the PR and it wasn't a complaint directed at you.
> 
> (I would also like to avoid potential doubts. I'm not a maintainer of the project)

No worries! :)

Hunter-Houts:
@mashb1t 
> Good changes, just a small improvement
Changes requested should be good to go now :) thank you for the suggestion!

mashb1t (approved):
lgtm 👍

lllyasviel (approved):

Issue #1924, opened 2024-01-14
clean up gradio temp files


hswlab:
**Is your feature request related to a problem? Please describe.**
I have noticed that the user interface also stores images in the ...\AppData\Local\Temp\gradio folder in addition to the Output folder. A lot of junk data accumulates there over time. 

**Describe the idea you'd like**
Is it possible that the temporary data is emptied after the image is created or that you can at least specify a custom path in the settings file that is easier to access for a manual clean up?

eddyizm:
@hswlab I think this is a OS thing, there's a ton of stuff that gets stored in temp if you use your computer a lot. checking mine it doesn't see like there is much but I tend to do weekly clean ups on my os. The outputs folder you can change that to your preferred location by updating the path in the config.txt file.

```
 [09:43]  eddyizm  ~
$ du -h AppData/Local/Temp/gradio/
0       AppData/Local/Temp/gradio/5d4d4a73540b9f44cc942466c44aa28b825242c1
0       AppData/Local/Temp/gradio/61e63f6acb5143c0ac1a2a630da880adc13b9fa9
0       AppData/Local/Temp/gradio/6368ac7336fbd1f2c5ad9bf589ff081bfed5f6af
0       AppData/Local/Temp/gradio/81447c3416325f610a8c5ebc35d1024f7a746c89
0       AppData/Local/Temp/gradio/bb01d9e06ab06e74c4befd031db121db2dc1c98e
0       AppData/Local/Temp/gradio/cbe968f3778826e2eabed6b81f27c0701121f1df
0       AppData/Local/Temp/gradio/e34df7a49b5584b871e07f7a889c39dbec868ee9
4.0K    AppData/Local/Temp/gradio/
```

mashb1t:
This heavily depends on the OS.
- Windows: Start > Settings > System > Storage >Toggle on Storage sense
- Linux: directory is cleared by default at every boot
- MacOS: 30 days cleanup period by default for untouched files (depending on last accessed)

FYI there is an arg called `--temp-path`, which currently (or rather until https://github.com/lllyasviel/Fooocus/pull/1932 is merged) does nothing as far as i can tell.

=> Fooocus should not mess with temp files of gradio, especially as also other gradio projects might be used on target system. There currently is no way of cleaning up or even accessing the paths of images yielded to gradio.

[...]

PR #1992, opened 2024-01-20
Feat: Add ability to change the Gradio temp path for saving images. Added ability to clear Gradio temp files and folders on launch

midareashi:
- Added option to change the Gradio Temp path
- Added option to delete Gradio Temp images and folders on launch

Changing the temp path is a feature requested by several users. I doubt a lot of users even know Gradio is filling up their appdata folder with temporary images each time they generate. In the future it would be nice to use the image generated in private_logger.py for the completed image and get rid of the duplicate images entirely. Until then...

Delete on launch will only happen when using a custom Gradio Temp path. That will stop this app from removing Gradio Temp images created by other applications.

EDIT @mashb1t: relates to https://github.com/lllyasviel/Fooocus/issues/1924


midareashi:
I would understand if you don't want to include the delete functionality. I just included it since it's what I have been using on my local to keep the temp files off my drive.

mashb1t:
relates to https://github.com/lllyasviel/Fooocus/pull/1932 where the file of private_logger.py is used as frontend output (but Gradio still somehow creates temp images, even when prividing a filepath)

mashb1t:
@midareashi can you please pull the current main branch into this PR and fix merge conflicts?

PR #2016, opened 2024-01-22
[enhance]: use opencv to rewrite morphological_open(), achieve a 40x speed increase in morphological_open opporation

rayleichenxi:
Hi~ 
     lllyasviel,
     
   Fooocus is  an excellent work, it can generate beautiful images in both text2img and img2img tasks. However, I encountered a speed bottleneck with Fooocus when using it for inpainting/outpainting tasks. The issue was with the `morphological_open` function, which was originally written by PIL, This implementation tends to become slow as the size of the input image increased. To address this, I rewrote the function using OpenCV, achieving a 40x increase in speed for image sizes below 3K in inpainting/outpainting tasks.

The following table presents a comparation of the execution speeds for the modified of `morphological_open`, which was tested on Intel(R) Xeon(R) Gold 6271C CPU @ 2.60GHz CPU and V100 GPU

image

### DIFF
--- a/modules/inpaint_worker.py
+++ b/modules/inpaint_worker.py
@@ -4,6 +4,7 @@ import numpy as np
 from PIL import Image, ImageFilter
 from modules.util import resample_image, set_image_shape_ceil, get_image_shape_ceil
 from modules.upscaler import perform_upscale
+import cv2
 
 
 inpaint_head_model = None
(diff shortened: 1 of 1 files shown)


poor7:
53x speed increase is amazing!

lllyasviel (approved):

PR #2038, opened 2024-01-25
chore: add venv mentioned in the readme to gitignore

Pradhyo:
This is for the [Linux (Using Python Venv)](https://github.com/lllyasviel/Fooocus?tab=readme-ov-file#linux-using-python-venv) section in the readme which has instructions to create the `fooocus_env` virtual environment. I updated `.gitignore` instead of the name in the readme for users who might have already followed these instructions.
 
```bash
python3 -m venv fooocus_env
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ user_path_config.txt
 user_path_config-deprecated.txt
 /modules/*.png
 /repositories
+/fooocus_env
 /venv
 /tmp
 /ui-config.json


mashb1t:
@Pradhyo which OS and which setup method are you using? Just to check if this is an individual problem with naming your venv or a general thing for all users on e.g. Linux (can't reproduce how you got the fooocus_env folder). Thanks for your input!

Pradhyo:
> @Pradhyo which OS and which setup method are you using? Just to check if this is an individual problem with naming your venv or a general thing for all users on e.g. Linux (can't reproduce how you got the fooocus_env folder). Thanks for your input!

Sorry @mashb1t, I just updated the description above.

lllyasviel (approved):

PR #2074, opened 2024-01-28
Bugfix/implement output path argument

eddyizm:


mashb1t:
@eddyizm starting to do reviews and merges now. I honestly think that always checking the path override when logging is not the most effective method. We could just check when loading the config if the path is overridden in args and then set the config instead.
What do you think?

eddyizm:
@mashb1t If I recall, I think I went this route because of having to compare a couple of different args, as they seem to be conflicting and felt it was easier to reason about and more robust in case of change mid use (even this is not yet currently possible). But yes, I can see it is not as efficient as checking it only once at start up. I can update the PR with that change.

mashb1t:
@eddyizm thank you for updating, much appreciated!

eddyizm:
ok, moved the function to the config, definitely cleaner. Tested locally and and on colab. 

What's interesting is even gradio fails to serve a directory outside of its root folder, you can always go into the folder and just open the html directly since the file/code is completely uncoupled, eg the html is independent. This works well in colab in particular since i mount my drive and then that drive is syncing to my desktop, I can just open up that window separately. 

May considering adding a note in the docs or even in the code/ui perhaps.

modules/config.py
```diff
@@ -102,6 +102,17 @@ def replace_config(old_key, new_key):
         print(e)
 
 
+def get_path_output() -> str:
+    ''' Checking output path argument and overriding default path
+        if disable image log path is not enabled.
+    '''
+    path_output = args_manager.args.temp_path if args_manager.args.disable_image_log else get_dir_or_set_default('path_outputs', '../outputs/')
```
mashb1t, line 109: Why do we need to check disable_image_log here and set the value to temp_path, due to get_current_html_path() in private_logger? Can't this just stay empty if disable_image_log is set?

mashb1t: or default to '../outputs/'?

modules/config.py
```diff
@@ -102,6 +102,17 @@ def replace_config(old_key, new_key):
         print(e)
 
 
+def get_path_output() -> str:
+    ''' Checking output path argument and overriding default path
+        if disable image log path is not enabled.
+    '''
```
mashb1t, line 108:
```suggestion
    """
    Checking output path argument and overriding default path if disable image log path is not enabled.
    """
```

modules/config.py
eddyizm: I think I had some concern originally when following `[--output-path OUTPUT_PATH] [--temp-path TEMP_PATH]` intended usages were. I could have sworn `temp_path` was being set to the local tmp directory when I initially wrote this PR but searching through the code now I don't see any reference to it which is odd. 
I guess I was trying to take into account when `--temp-path` would be passed in but looking at it now, I don't see that being used or checked anywhere either. 

I believe overall you are correct as I can't recall what I was thinking 2 weeks ago. :-)

eddyizm: went and ahead and removed that check, left the if statement so we can see it in the log when it is overwritten.

PR #2102, opened 2024-01-30
Wildprompt

midareashi:
Wildprompt lets you organize sets of prompts in files. You can choose specific files to pull random prompts from and generate images. It supports multiple file selection. For instance, combining Sci-Fi and Fantasy prompts can create a mash up of fantasy Sci-Fi images. It also works with wildcard to increase the variety of images generated.

Includes a Generate All option which will create an image for each prompt in the selected file. Increase the Image Number to increase the number of times each prompt is generated. Only works if one prompt file is selected because the number of images generated would grow out of control.

This also updates the output folder to create sub-folders named with the wildprompts used in the batch. This helps organize the output folder when generating lots of images.

image


midareashi:
Added feature to filter the private log by base model name and wildprompt names.
image

DavidKouw:
Hi there! Any new about implementing this in new version of Foo?

Issue #2126, opened 2024-02-01
Add Authentication also for --listen API and add documentation


Schm1tz1:
**Is your feature request related to a problem? Please describe.**
In many cases you don't want unauthenticated access to the UI even in `--listen` mode. Assuming that you have some kind of forwarding/NAT/reverse proxy in place the interface might be secured and / or encrypted with TLS (e.b SSH-Tunnel, NGINX/HAPROXY/Caddy or a cloudflare tunnel) but still unauthenticated access IS possible.

**Describe the idea you'd like**
Allow authentication not only for `--share` but also for `--listen` and add a few lines of docs how to use this feature (so far there's only an example). 

Implementation is a one-liner I have done locally and will share as a PR. To be tested in various scenarios.

Schm1tz1:
It's basically removing the check for share that will use auth in both scenarios:
https://gist.github.com/Schm1tz1/4fbd0e24035cf19d25e8926e9e59dd69#file-webui-py-L621

Schm1tz1:
PR was tested in all 4 possible scenarios: with/without AuthN for both --share and --listen

PR #2127, opened 2024-02-01
Update webui.py

Schm1tz1:
Enabling authentication for local and shared UI (--listen and --share), see https://github.com/lllyasviel/Fooocus/issues/2126

### DIFF
--- a/webui.py
+++ b/webui.py
@@ -618,6 +618,6 @@ shared.gradio_root.launch(
     server_name=args_manager.args.listen,
     server_port=args_manager.args.port,
     share=args_manager.args.share,
-    auth=check_auth if args_manager.args.share and auth_enabled else None,
+    auth=check_auth if auth_enabled else None,
     blocked_paths=[constants.AUTH_FILENAME]
 )
--- a/webui.py
+++ b/webui.py
@@ -618,6 +618,6 @@ shared.gradio_root.launch(
     server_name=args_manager.args.listen,
     server_port=args_manager.args.port,
     share=args_manager.args.share,
-    auth=check_auth if auth_enabled else None,
+    auth=check_auth if (args_manager.args.share or args_manager.args.listen) and auth_enabled else None,
     blocked_paths=[constants.AUTH_FILENAME]
 )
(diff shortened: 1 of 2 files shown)


mashb1t:
Some other users and me have only recently chatted about this exact feature somewhere in the discussions 👍

Schm1tz1:
NIce! I am using a reverse proxy to access Fooocus from the outside and are testing with Cloudflare Tunnels and I didn't want unauthenticated access so I tried this modification. Seems to be backwards-compatible, non-breaking.

webui.py
```diff
@@ -618,6 +618,6 @@ def dump_default_english_config():
     server_name=args_manager.args.listen,
     server_port=args_manager.args.port,
     share=args_manager.args.share,
-    auth=check_auth if args_manager.args.share and auth_enabled else None,
+    auth=check_auth if auth_enabled else None,
```
mashb1t, line 621:
```suggestion
    auth=check_auth if (args_manager.args.share or args_manager.args.listen) and auth_enabled else None,
```


I don't see any reason why we should only restrict this for users using --share and not also allow basic auth protection for --listen. For this you normally would also want to rather add args_manager.args.listen than also using basic auth locally. Is there any reason why you'd also propose to have basic auth locally?

mashb1t (changes requested):

mashb1t:
> NIce! I am using a reverse proxy to access Fooocus from the outside and are testing with Cloudflare Tunnels and I didn't want unauthenticated access so I tried this modification. Seems to be backwards-compatible, non-breaking.

IMHO the best case scenario in this case would be to not even route unauthenticated traffic to Fooocus at all but filter it in the reverse proxy.

webui.py
Schm1tz1: Fair point - didn't have the local mode on screen but why not allow authentication in general and let the user decide, that's what I did here. You still actively need to create the `auth.json` file so no surprises are to be expected.

Schm1tz1:
> > NIce! I am using a reverse proxy to access Fooocus from the outside and are testing with Cloudflare Tunnels and I didn't want unauthenticated access so I tried this modification. Seems to be backwards-compatible, non-breaking.
> 
> IMHO the best case scenario in this case would be to not even route unauthenticated traffic to Fooocus at all but filter it in the reverse proxy.

For that case it makes senes but let's not discuss that here - it depends a lot on the whole setup and how your Ingress / Proxy / NAT setup looks like. It's just about allowing basic authentication on Fooocus - one can still decide to move it or not.

Issue #2131, opened 2024-02-01
The bug in the FaceSwap feature.


diaolulu1:
The FaceSwap function switches models at the 15th step, after which the subsequent refinement makes the faces, which were originally quite similar, look completely dissimilar.

mashb1t:
It would be great if you could investigate if this is a 50% or 15th step issue (changing to quality and checking again) and if you could pin point the code line stating the switch. Thank you!

diaolulu1:
I think it should be caused by the Refiner feature; the default Refiner Switch is set to activate at 50%. As long as I don't enable the Refiner, there's no problem.
![01](https://github.com/lllyasviel/Fooocus/assets/55993721/fbeb292a-0c90-4ab6-b02c-df5ed402e3a7)
![02](https://github.com/lllyasviel/Fooocus/assets/55993721/7d064b5a-0a42-4e6a-b1c0-468c2c2e549f)
![03](https://github.com/lllyasviel/Fooocus/assets/55993721/b66851f0-b9fd-41ca-90e8-a50e7e17697e)
![04](https://github.com/lllyasviel/Fooocus/assets/55993721/2f81a37e-3ddd-4b79-ab93-eb85a48f79d8)

xhoxye:
https://github.com/lllyasviel/Fooocus/blob/1bcbd6501bb3e5db5674a18a8a8074034cda318b/modules/async_worker.py#L307
It is indeed 0.5 of the lock, which will cause all guidance to lose control.

dav30383:
I still have the same problem, and I've never used ReFiner! For a certain point it seems to follow the right face, and then it changes it! But wouldn't it be possible to implement ReActor??? Much more effective when it comes to face swapping!

mashb1t:
This will be reevaluated, for now we have https://github.com/lllyasviel/Fooocus/pull/2165

dav30383:
Sorry, but why don't you fix it now instead? It's a serious problem! I've already given you the solution, just implement the ReActor code or add it as a feature, right?

mashb1t:
@dav30383 please calm down. See https://github.com/lllyasviel/Fooocus/discussions/2154#discussioncomment-8389770, i'll focus on bugfixes first before adding new geazures, especially sich which need external packages / libraries like insightface or similar.

LIQUIDMIND111:
[...]

PR #2165, opened 2024-02-05
fix refiner switching issues

xhoxye:
https://github.com/lllyasviel/Fooocus/issues/2131

1. In inpaint, when refiner is None, enable use_synthetic_refiner. The default switching timing of 0.5 is too early, which is now modified to SDXL default of 0.8.
2. When using custom steps, the calculation of switching timing is wrong. Now it is modified to calculate "steps x timing" after custom steps are used.


mashb1t (approved):

modules/async_worker.py
```diff
@@ -304,7 +304,7 @@ def handler(async_task):
                         print(f'[Inpaint] Current inpaint model is {inpaint_patch_model_path}')
                         if refiner_model_name == 'None':
                             use_synthetic_refiner = True
-                            refiner_switch = 0.5
+                            refiner_switch = 0.8
```
mashb1t, line 307: @xhoxye can you please elaborate why exactly 0.8 should be the new value here?

xhoxye: 0.8 is the default switching timing value for SDXL refiner, which reduces the loss of control of the picture to a small effect, of course 1.0 is OK, but 1.0 will turn off use_synthetic_refiner

mashb1t: I'm uncertain what the purpose of using use_synthetic_refiner is. Does it allow for more unguided steps? But this is what the ControlNet stop at is for anyways, right?

mashb1t: i've tested both variants, and for me basically nothing changes when setting refiner_switch to 0.5 or 0.8 without an actual refiner. 

Prompt: "a man wearing a red suit"
Seed: 0
Resolution: 1152×896 (default)
Stop At: 0.9 (default)
Weight: 0.75 (default)

Base image:
base

0.5:
image - 2024-02-11T134154 584

0.8:
image - 2024-02-11T140123 006

I'd merge the fix for refiner switch calculation but not for the first part until clarified further.
Is there something i'm missing?

xhoxye:
Can you get another Asian face. Try it.
it becomes a European after 0.5.
QQ截图20240211213714
0.5

mashb1t:
yes, got it. Updated comment https://github.com/lllyasviel/Fooocus/pull/2165#discussion_r1485592011

modules/async_worker.py
xhoxye: Can you get another Asian face. Try it.
it becomes a European after 0.5 or 0.7.
QQ截图20240211214143

image

mashb1t:
> Could it be that Asians face and models didn't match?

This could be possible, still switching to refiner, even synthetic one, will again introduce noise depending on switch step and cause the result to not apply faceswap / inpaint correctly

mashb1t:
@lllyasviel IMHO this is a non-intended side-effect but it's also possible that we're not getting the full picture here.
Is it intended that refiner_switch is 0.5 and not 1 when no refiner is set and if yes, what are the reasons?

https://github.com/lllyasviel/Fooocus/blob/2.1.864/modules/async_worker.py#L307

mashb1t:
@lllyasviel your feedback would be highly appreciated here to identify a potential bug and to understand the implementation better.

ChineseNorris:
I've checked a few things with the refiner_switch = 0.8 and here are some issues I have found:

- With refiner_switch = 0.8 the image gets more saturated and looks a little bit "better" but the inpaint(masked) creates a thin white line that looks like it's glowing
- With refiner_switch = 0.5 the image looks more natural and the inpaint doesn't do weird things
- With refiner_switch = 1 the saturation is more clear and doesn't look as natural as 0.5

Maybe change refiner_switch to be altered in profiles and change this setting if anyone needs it?
refiner_switch = 0.8

refiner_switch = 0.5

PR #2212, opened 2024-02-08
Add sensitive case handler

MindOfMatter:
Added a fix for get_config_item_or_set_default to better handle the case-sensitivity problem (don't get loras with a typo, style or model problem).

```
def correct_case_sensitivity(value, valid_values):
    """Corrects case sensitivity of a value or list of values based on a list of valid values.
       If a valid value is contained in the input value, it replaces it with the full valid value."""
     ...
     
by example
     
def sdxl_styles_corrector(value):
    return correct_case_sensitivity(value, modules.sdxl_styles.legal_style_names)

default_styles = get_config_item_or_set_default(
    key='default_styles',
    default_value=[
        "Fooocus V2",
        "Fooocus Enhance",
        "Fooocus Sharp"
    ],
    validator=sdxl_styles_validator,
    corrector=sdxl_styles_corrector
)
```

so if we have "My Hyper realistic preference Negative" it will fit with "My Hyper Realistic Preference Negative" in default_styles configuration

My changes :

- replace all validator with better functions
- add new corrector function (if validator first fail)
- add new corrector function (if validator first fail)
- add flags debug logs in function (optional)
- improve get_config_item_or_set_default to handle each cases
- add better validation for get_model_filenames
- generic correct_case_sensitivity handle all possible issue cases
- improve the loras_validator
- add better sdxl_styles_validator / corrector (not have missing expected value anymore)
- improve the error logs cases
- change update_all_model_names call oder to better verification


README.md
```diff
@@ -0,0 +1,39 @@
+# Fooocus-MindOfMatter-Edition
```
mashb1t, line 1: @MindOfMatter just to make sure we're getting the baseline straight: please remove your readme changes from the PRs as Fooocus is not planned to be rebranded to Fooocus-MindOfMatter-Edition any time soon ^^
Thank you!

mashb1t (changes requested):

PR #2258, opened 2024-02-14
Settings cleanup

docppp:
This commit in an attempt to refactor config file with the power of attrs library. With this changes, any additional settings are easy to create; user can easily configure every option; presets is parsed same as config.txt; invalid options provided with confix.txt results in fallback to hardcoded in-code defaults.


mashb1t:
@docppp appreciate the effort, thank you very much! There have been a few releases not included in your PR, plus some functionalities are missing, e.g. folder creation if dir does not exist and fallback to default happens as well as output path handling with args.
Please consider updating this PR to do enable code evaluation.

PR #2262, opened 2024-02-14
NO-STORY: Remove skip & stop buttons

hgleyder:


mashb1t:
Why would you want to not show the skip and stop button?

PR #2313, opened 2024-02-21
Ignore DS_Store

charliewilco:
This pull request updates the .gitignore file to exclude .DS_Store files from being tracked by Git. .DS_Store stands for Desktop Services Store, and these files are automatically created by macOS to store custom attributes of a folder such as the position of icons or the choice of a background image.


mashb1t:
@charliewilco i personall have the .DS_Store ignored in the global .gitignore as this affects all projects and you generally don't ever want to have the conmitted, right?

PR #2323, opened 2024-02-22
Adding support to inline prompt references

cantor-set:
## Objective:

* To allow to specify LoRAS using the prompt string.

## Implementation:

* Removed circular reference on utils.py by moving the function `get_files_from_folder` to another file.
* Removed circular reference on modules/sdxl_styles.py by moving the function `apply_wildcards` to `modules.config`.
* Adds a function that parses prompt tokens, assuming comma delimited entries. 
* A regular expression captures the groups for a Lora name and weight
* Adds an initial md meant to be used as a guideline for development
* Added a `tests` directory

### Limitations
* LoRAS references in wildcards strings are not supported

## Usage
```
key_word1, keyword2, <lora:some-lora:1>, <lora:another-lora:0.34>
```


cantor-set:
@mashb1t Inputs are welcome

mashb1t:
@cantor-set Fooocus aims for simplicity and ease of use. and LoRA activation can already be done using the select inputs. What's your reasoning for the LoRA reading functionality from prompt strings?

cantor-set:
> @cantor-set Fooocus aims for simplicity and ease of use. and LoRA activation can already be done using the select inputs. What's your reasoning for the LoRA reading functionality from prompt strings?

The reasoning is convenience for advance users. Faster adjustment and setting of multiple LoRAs without having to swap windows or clicking away to search and drop down menu selection,  minimizing the number of steps to accomplish the same objective. Specially when the user knows which LoRa needs to be applied. 

Additionally, it does not interferes with the normal flow of operations, it just adds more options for advance users.

ujl123:
> @cantor-set Fooocus aims for simplicity and ease of use. and LoRA activation can already be done using the select inputs. What's your reasoning for the LoRA reading functionality from prompt strings?

Another issue is that "slider" loras can often use numbers greater than 2 and lower than -2. 

https://civitai.com/models/112552/weight-slider-lora -3 to 3
https://civitai.com/models/228180/diablo-sliders-ntcaixyz examples shown uses 5

It also allows for wildcards to use loras.

mashb1t:
@cantor-set @ujl123 thank you for the feedback, i can see why this can be useful.
Feel free to finalize the implementation and mark it as ready to review.

cantor-set:
@mashb1t  @lllyasviel Ready for review and suggestions.

mashb1t:
@cantor-set please check again with hte latest version 2.2.0 and pull it to your branch.

cantor-set:
> @cantor-set please check again with the latest version 2.2.0 and pull it to your branch.
@mashb1t Done. Added an additional change that fixes a circular dependency on `utils.py`.

modules/async_worker.py
```diff
@@ -426,6 +426,10 @@ def handler(async_task):
             extra_negative_prompts = negative_prompts[1:] if len(negative_prompts) > 1 else []
 
             progressbar(async_task, 3, 'Loading models ...')
+
+            # Parse lora references from prompt
+            loras = parse_lora_references_from_prompt(prompt, loras, loras_limit=modules.config.default_max_lora_number)
```
mashb1t, line 431:
```suggestion
            loras = parse_lora_references_from_prompt(prompt, loras, modules.config.default_max_lora_number)
```

+ don't you agree we should also replace the lora notation in the prompt for correct metadata handling, logging etc.?

mashb1t (changes requested):

modules/util.py
```diff
@@ -14,8 +15,16 @@
 import modules.sdxl_styles
 
 LANCZOS = (Image.Resampling.LANCZOS if hasattr(Image, 'Resampling') else Image.LANCZOS)
+
+
+# Regexp compiled once. Matches entries with the following pattern:
+# <lora:some_lora:1>
+# <lora:aNotherLora:-1.6>
+LORAS_PROMPT_PATTERN = re.compile(".*<lora:(.+):([-+]?(?:\d*\.*\d*))>.*")
```
chocolateboy, line 23: This incorrectly matches e.g. `<lora:foo:1..2>`, `<lora:bar:.>`, `<lora:baz:+>` and `<lora:quux:>`. Also, some backtracking can be avoided, e.g.:

```python
LORA_PROMPT_PATTERN = re.compile(r".* <lora : ([^:]+) : ([+-]? (?: (?:\d+ (?:\.\d*)?) | (?:\.\d+)))> .*", re.X)
```

This matches, e.g.:

- `<lora:foo:1>`
- `<lora:bar:-1.2>`
- `<lora:baz:1.>`
- `<lora:quux:+.2>`

cantor-set: I will make the change and include those cases in the unit test. Thanks @chocolateboy

mashb1t (approved):

PR #1940, opened 2024-01-15
feat: add metadata to images

mashb1t:
Resolves https://github.com/lllyasviel/Fooocus/issues/241 + #1996
Merge https://github.com/lllyasviel/Fooocus/pull/1932 to also enable metadata in images in the browser, not only images in the outputs folder.

This feature is in contrast to https://github.com/lllyasviel/Fooocus/discussions/160 and might not be merged, but at least if offers the functionality to whoever wants to use it locally, as there was demand for it.

I've implemented both a Fooocus (json) and A1111 (plain text) meta data scheme, where the latter is 100% compatible with A1111 and Civitai, but can not be used to reproduce the image outside of Fooocus, as there are so many improvements and special things happening in Fooocus it's just not applicable anywhere else.

The A1111 meta data scheme is based on [A1111 infotext handling](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/cf2772fab0af5573da775e7437e6acdca424f26e/modules/processing.py#L672-L700) and i've re-implemented the style resolution feature to extract styles and check the corresponding styles checkboxes.

Both schemes might be extended in the future.

This feature also introduces various code optimisations such as extraction of Performance and Steps enums as well as centralisation of values for ControlNet count and LoRA count.

<details><summary>Gradio (setting in Developer Debug Mode)</summary>
<p>

Default is Fooocus Scheme
image

</p>
</details> 

<details><summary>Config options</summary>
<p>

    "default_save_metadata_to_images": true,
    "default_metadata_scheme": "a1111",
    "metadata_created_by": "mashb1t"

</p>
</details> 

<details><summary>Arg --disable-metadata</summary>
<p>

 `--disable-metadata` completely prevents metadata processing and output in Gradio

</p>
</details> 

<details><summary>Metadata Reader</summary>
<p>

1. open Image Input > Metadata tab
2. drag & Drop image to image upload
3. automatic preview of image metadata
4. apply metadata to Gradio inputs on button click

Fooocus scheme
Screenshot 2024-01-29 at 15 13 17

A1111 scheme
Screenshot 2024-01-29 at 15 09 52

</p>
</details> 

<details><summary>Metadata in files</summary>
<p>

Speed Fooocus scheme
image

LCM A1111 scheme (yes, with negative prompt, because it technically exists but doesn't have an influence)
image

Speed A1111 scheme
image

</p>
</details> 

<details><summary>Civitai</summary>
<p>

Speed Fooocus scheme
image

LCM A1111 scheme
image

image

[...]

PR #2425, opened 2024-03-03
feat: add metadata hint to history log

mashb1t:
As discussed in https://github.com/lllyasviel/Fooocus/pull/1940#issuecomment-1975123101

image

### DIFF
--- a/modules/async_worker.py
+++ b/modules/async_worker.py
@@ -857,17 +857,17 @@ def worker():
                     if freeu_enabled:
                         d.append(('FreeU', 'freeu', str((freeu_b1, freeu_b2, freeu_s1, freeu_s2))))
 
+                    for li, (n, w) in enumerate(loras):
+                        if n != 'None':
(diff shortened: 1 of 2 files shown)


poor7:
Add also **Refiner swap method** if it is not equal to **joint** and Forced **Overwrite of Sampling Step**, **Forced Overwrite of Refiner Switch Step** if they are not equal to **-1** or greater than **0**.

mashb1t:
> The saving method also needs to be specified, because there are 2 of them.
> 
> Add also **Refiner swap method** if it is not equal to **joint** and Forced **Overwrite of Sampling Step**, **Forced Overwrite of Refiner Switch Step** if they are not equal to **-1** or greater than **0**.

I've added steps on override to history log, the rest is already handled correctly. Please check out https://github.com/lllyasviel/Fooocus/blob/main/modules/async_worker.py#L845-L851

Benm678:
@mashb1t - Mate! You guys are awesome, thank you!

PR #2466, opened 2024-03-06
feat: Scan wildcard subdirectories

### LINKED ISSUE #2461: [Feature Request]: Scan subdirectories for wildcards
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits

### What would your feature do?

When working with a large number of different wildcards, it gets difficult tracking down the correct file sometimes. When working with composite wildcards (wildcards referencing other wildcards), you tend to end up with a lot of them. I think it would be beneficial to allow users to sort their wildcards in folders, as it's often accessed by users, and doesn't have a front-end representation.

### Proposed workflow

1. Create subdirectories and wildcards
2. Fooocus automatically scans all subdirectories for wildcards

### Additional information

If this is a desirable feature, I would be willing to look into implementing it.

Cruxial0:
Closes #2461 

Allows users to sort their wildcards in folders. Previously Fooocus would only search the base directory.

- Implements functionality for scanning subdirectories within the wildcards directory
- Adds a function for returning the first occurrence of a file within a directory tree

### DIFF
--- a/modules/sdxl_styles.py
+++ b/modules/sdxl_styles.py
@@ -6,10 +6,8 @@ import modules.config
 
 from modules.util import get_files_from_folder
 
-
 # cannot use modules.config - validators causing circular imports
 styles_path = os.path.abspath(os.path.join(os.path.dirname(__file__), '../sdxl_styles/'))
-wildcards_path = os.path.abspath(os.path.join(os.path.dirname(__file__), '../wildcards/'))
 wildcards_max_bfs_depth = 64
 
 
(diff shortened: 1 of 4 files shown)


modules/sdxl_styles.py
```diff
@@ -3,7 +3,8 @@
 import json
 import math
 
-from modules.util import get_files_from_folder
+from modules.util import get_files_from_folder, get_first_occurrence_in_folder
+from glob import glob
```
mashb1t, line 7: where is glob used in the code?

modules/util.py
```diff
@@ -175,13 +175,23 @@ def get_files_from_folder(folder_path, exensions=None, name_filter=None):
...
             _, file_extension = os.path.splitext(filename)
-            if (exensions is None or file_extension.lower() in exensions) and (name_filter is None or name_filter in _):
+            if (extensions is None or file_extension.lower() in extensions) and (name_filter is None or name_filter in _):
                 path = os.path.join(relative_path, filename)
                 filenames.append(path)
 
     return filenames
 
 
+def get_first_occurrence_in_folder(folder_path, filename):
```
mashb1t, line 185: please check out the already existing model / lora file functions, which might be reusable here

mashb1t (changes requested):

MichalZak17 (approved):

mashb1t (approved):
@Cruxial0 i've added the ability to configure wildcards filepaths incl. caching as the wildcard filepaths were collected every time a wildcard was processed (in the max. 64 iterations for loop).
Please double check so we can merge this feature to develop for the next release.

Cruxial0:
> Please double check so we can merge this feature to develop for the next release.

Looks good, however I'm not sure how much I like having to refresh files every time a wildcard is added. When I work with wildcards, I tend to add new files frequently. It's by no means a deal-breaker, but I believe it would be more user friendly to update it automatically. Maybe a function that compares the number of files in the wildcard directory at the start of every generation, and only updates if files were added/removed?

mashb1t:
@Cruxial0 the options are 
- cache content of wildcards, access cached array only => least flexible, but most performant. Downside: linear increase in memory consumption depending on file content of wildcards
- loading files in apply_wildcards => loaded for each image number iteration
- same handling as styles_files => only loaded on application start
- method called before apply_wildcards outside the task loop => list needs to be collected for every generation click
- store file list in config, reload with reload files button =>  one time file list processing, allows dynamic content changes while still offering reload without having to restart Fooocus

The changes of this MR now also allow nested directory structures and recursive collection, which is why file counting is not going to work, also one file can be deleted and another added, which is why a simple count isn't sufficient here.

I think the currently implemented option offers the best compromise between performance and flexibility.

PR #2519, opened 2024-03-13
Update xformers to 0.0.23 in Dockerfile

josephrocca:
To match this: https://github.com/lllyasviel/Fooocus/pull/2517

Just tested on main branch with this change, and all works fine.


mashb1t:
Thanks, much appreciated!

PR #2590, opened 2024-03-21
Slightly more compact ui settings

Alexdnk:
Changed Radio to Dropdown. If there are a lot of presets, it's a ui problem with Radiobutton.
And i suggest hiding the resolution selection under Accordion.
And simplify the file type selection.


xhoxye:
Presets can be pulled down because users may create an indefinite number of presets.

JulienZLQ:
> Fooocus follows the philisophy of always showing available options, not hiding them behind a select or accordion. While i get the intention of yours, we still should keep this in mind.

I personally can understand the original intention of fooocus, but based on the native fooocus model drop-down options, can it be understood that due to the uncertain number of models, the use of drop-downs is allowed to avoid excessive stacking of options.
If so, other unopened customizations can remain as they are, but it is indeed preset, which has the same uncertainty as the model and can be configured a lot. Will this part be adjusted to a drop-down?

mashb1t:
Will be reevaluated in a few weeks, background info see https://github.com/lllyasviel/Fooocus/discussions/2617#discussioncomment-8888224

HelloLaohe:
> Fooocus 遵循的理念是始终显示可用选项，而不是将它们隐藏在选择或手风琴后面。 虽然我明白你的意图，但我们仍然应该记住这一点。
I fully support this viewpoint, and believe that intuitiveness is the most important factor. I hope that the style preview images can be displayed directly, without the need for mouse hovering to make them appear.

mashb1t:
Proposal 1: add select for presets
screencapture-localhost-7865-2024-05-23-01_10_52

Proposal 2: add select for presets + accordion for resolutions, but fixed to 2 column layout
screencapture-localhost-7865-2024-05-23-01_10_31

screencapture-localhost-7865-2024-05-23-01_17_18

Your feedback is welcome!

Alexdnk:
Thank you for the review! Please forgive me I am a novice programmer and I am having a hard time understanding code and using GitHub in general. 
I suggest removing the word 'Resolution', since it wasn't in the original interface.
```
with gr.Accordion(label='Aspect Ratios   width × height', open=False) as aspect_ratios_accordion:
	aspect_ratios_selection = gr.Radio(label='Aspect Ratios', show_label=False, choices=modules.config.available_aspect_ratios,
									   value=modules.config.default_aspect_ratio,
									   elem_classes='aspect_ratios')
```
Снимок экрана 2024-05-23 094016

It may also be useful to add a line with information about the selected resolution. However, in my current implementation it is updated only when any resolution item is selected. At startup it is empty and shows nothing. I don't know how to fix this or if it is necessary. 
I would like to place this information line in the accordion as I marked in red on the screenshot. But I don't know how.

```
aspect_ratios_markdown = gr.Markdown('')
def aspect_ratios_change(value):
	return value
aspect_ratios_selection.change(aspect_ratios_change, inputs=aspect_ratios_selection, outputs=aspect_ratios_markdown,
			   queue=False, show_progress=False)
```
Снимок экрана 2024-05-23 094114

mashb1t:
@Alexdnk adjusted, please check

Alexdnk:
@mashb1t   'Aspect Ratios' Accordion it's is great realisation. I test it, and is ok.
In the 'Output Format' section, did you intentionally swap the format and quantity? In my original pullrequest, I suggested combining them into one line(Row), so I have them in that order. 
I suggest not to change the old order and to place the 'Image Number' above the 'Output Format'.
But if that's the way you intend it to work, then it works correctly.
Thank you for adjusting.

PR #2606, opened 2024-03-22
prevent of potential type's mismatch

igeni:
replaced concatenation by the f-strings to avoid potential type's mismatch and make code more clear


mashb1t:
closing as type is always string
https://github.com/lllyasviel/Fooocus/blob/e2f9bcb11d06216d6800676c48d8d74d6fd77a4b/extras/BLIP/models/blip_pretrain.py#L94

PR #2614, opened 2024-03-23
Updated git properties to ignore wildcard/presets

arakinas:


arakinas:
Made updates to the speed settings and inpainting to allow for more default speeds, and less inpainting effort.

PR #2805, opened 2024-04-24
feat: build and push container image for ghcr.io, update docker.md, and other related fixes 

xynydev:
I had some issues with the container and wanted to contrib the fixes back upstream. This has been tested to work on Fedora 39 and in GitHub Actions. Publishing an image on ghcr.io has no extra cost to you as the maintainer, but will make the life of users who prefer containers much easier.

See commits for more details. I'm also happy to answer questions.


mashb1t:
Hey @xynydev, thank you very much for your effort, much appreciated!
The thing is: 
The initial Docker MR alrwady included external regustry publishing, but only @lllyasviel can add secrets to this repository, and as he hasn't responded since we've added Docker support we'd need him to set Secrets up in the first place...
So we're currently stuck here.

.github/workflows/build_container.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/${{ github.repository_owner }}/${{ env.IMAGE_NAME }}
```
mashb1t, line 35:
```suggestion
          images: ghcr.io/${{ github.repository_owner }}/${{ github.event.repository.name }}
```

.github/workflows/build_container.yml
```diff
@@ -0,0 +1,48 @@
+name: Create and publish a container image
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+env:
+  IMAGE_NAME: fooocus
```
mashb1t, line 10:
```suggestion
```

.github/workflows/build_container.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/${{ github.repository_owner }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=sha
```
mashb1t, line 39:
```suggestion
            type=semver,pattern={{version}}
            type=semver,pattern={{major}}.{{minor}}
            type=semver,pattern={{major}}
```

.github/workflows/build_container.yml
```diff
@@ -0,0 +1,48 @@
+name: Create and publish a container image
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
mashb1t, line 7:
```suggestion
  push:
    tags:
      - 'v*'
```

entrypoint.sh
```diff
@@ -30,4 +30,4 @@ mklink outputs
 import outputs
 
 # Start application
-python launch.py $*
+python entry_with_update.py $*
```
mashb1t, line 33: users normally do not expect Fooocus to be suddenly updated when running a container tagged with a specific version

```suggestion
python launch.py $*
```

docker.md
```diff
@@ -1,35 +1,99 @@
...
+
+Build with `podman`:
+```sh
+podman build . -t fooocus
+```
 
 ## Details
 
-### Update the container manually
+### Update the container manually (`docker-compose`)
```
mashb1t, line 82:
```suggestion
### Update the container manually (`docker compose`)
```

mashb1t (changes requested):

mashb1t:
Hey @xynydev, I've actually never worked with Github actions to ghcr.io before, so thanks a lot for the explanation.
The goal is only to build an image on releases tagged with versions (semver conform) and I'll have to do the switch from X.X.X to vX.X.X scheme to better distinguish version tags from other ones.
Please double-check the changes.

.github/workflows/build_container.yml
xynydev: GHCR supports only lowercase image names unfortunately, and the repository name is uppercase. I'll merge this for now, as I believe the docker action should make the inputs lowercase automatically, but if not, we're going to have to revert this change.

xynydev:
> The goal is only to build an image on releases tagged with versions (semver conform) and I'll have to do the switch from X.X.X to vX.X.X scheme to better distinguish version tags from other ones.
Please double-check the changes.

Sounds great. I went ahead and merged your changes. I haven't used the semver option in the docker metadata action, but your suggestion looks alright. I'll see if I can do testing in my repo to validate that it works.

xynydev:
Aight, it worked: 
https://github.com/xynydev/Fooocus-fork/actions/runs/9194578288/job/25288353110
https://github.com/xynydev/Fooocus-fork/pkgs/container/fooocus-fork

The tag is `2.3.1` not `v2.3.1`, though. It might be possible to override that somehow, but everything works well enough now I guess.

Note that the current `on: push:` condition will only create a new container image when a new tag is created with a new commit in it. This would _probably_ mean that the previous versions wouldn't get container images, only the next one. Also, if anyone has a reason to use an older image, they will get the one that was built that one time, without any fixes that might have been made upstream or in Fooocus' Dockerfile.

mashb1t:
@xynydev thanks for testing, much appreciated!
Tagging in docker with 2.3.1 is expected, git tag has to be v2.3.1 though.

> Also, if anyone has a reason to use an older image, they will get the one that was built that one time, without any fixes that might have been made upstream or in Fooocus' Dockerfile.

This is also expected.
Older images are not necessarily planned to be built, first one is the upcoming 2.4.0.
Do you know if a tag on an existing commit will also trigger the pipeline? (this would be the preferred behavior)

xynydev:
From my testing, I'm not sure. I first created a tag on an existing commit and nothing happened. I then made a meaningles commit, pushed it, tagged it locally, and pushed the tag, and then it was triggered.

mashb1t:
@xynydev just tested, the tag has to be on a branch where this change is already included, see https://github.com/mashb1t/Fooocus/actions/runs/9196855992/job/25296038248 
Tagging and building older versions is consequently not possible, but also not needed.

mashb1t:
I assume this is a personal issue but I'm getting this error with driver 555.85:
RuntimeError: Unexpected error from cudaGetDeviceCount(). Did you run some cuda functions before calling NumCudaDevices() that might have already set an error? Error 500: named symbol not found

May i ask which driver you're using?

mashb1t (approved):

PR #2926, opened 2024-05-15
feat: UX Colorize aspect ratios, landscape vs portrait

zeonglow:
I found trying to guess if the Aspect Ration was landscape or portrait. Kept getting it wrong! As my computer is super old and slow, this bugged me.

My original idea was to split the manual in three, but that was a hassle to implement without changing a lot of things. This seemed pretty straight forward.

I'm so lazy I didn't add a key, but that would just create more work for localization and it's pretty clear

Screenshot
image


andynormancx:
It feels to me that this is crying out for a little bit of SVG magic to draw the aspect ratio next to each option…

PR #2936, opened 2024-05-17
Let gallery start in preview mode [video]

MohammadAboulEla:
Gallery will start in preview mode, which shows all of the images as thumbnails and allows the user to click on them to view them in full size, this will be the logical option when user generate 1 image at time so the generated image will not be minimized for no reason after being generated.  when generated multiple images and since the user will activate it any way to see his images it is also logical to activate it after images being generated.
see attached video 
https://github.com/lllyasviel/Fooocus/assets/89942942/397d8d39-0f80-4f84-92fb-a4d1b220047c

### DIFF
--- a/webui.py
+++ b/webui.py
@@ -111,7 +111,7 @@ with shared.gradio_root:
                                     elem_id='progress-bar', elem_classes='progress-bar')
             gallery = gr.Gallery(label='Gallery', show_label=False, object_fit='contain', visible=True, height=768,
                                  elem_classes=['resizable_area', 'main_view', 'final_gallery', 'image_gallery'],
-                                 elem_id='final_gallery')
+                                 elem_id='final_gallery', preview=True)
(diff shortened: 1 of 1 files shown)


mashb1t:
Hey @MohammadAboulEla,
thank you for your contribution. This is a subjective topic, in my opinion from a user flow perspective the user should start with large previews to better be able to select the generated images, then go into detaill and see even more. This is also aligned with the intermediate image gallery output while generating (on the right of the preview image).
It is absolutely valid to keep the preview in your fork and to have  a different opinion about it, but for now it's a "won't change".

mashb1t:
@MohammadAboulEla as this is a setting on the gallery element itself it would be global or needs to be updated on every generation, which would not be consistent then.

kalle07:
iam with you [MohammadAboulEla](https://github.com/MohammadAboulEla) ... but it doesn't seem so easy to treat one picture differently than several

PR #2943, opened 2024-05-17
feature: update anime preset

mashb1t:


SP-l33t:
@lllyasviel It works, can approve 😁

PR #2968, opened 2024-05-20
feat: only use valid inline loras, add subfolder support

mashb1t:


ErroneousBosch:
if I have a lora in `loras/foo/bar.safetensors`, would that be `<lora:bar:1>` or `<lora:foo/bar:1>`?

mashb1t:
@ErroneousBosch this would be `<lora:bar:1>`
as the file path stem is used, see https://github.com/lllyasviel/Fooocus/blob/main/modules/util.py#L413

PR #2971, opened 2024-05-21
feature: Reads the size and ratio of the image and gives the recommended size

xhoxye:
QQ截图20240521085617


webui.py
```diff
@@ -220,7 +224,47 @@ def ip_advance_checked(x):
...
+                                            lcm_ratio = f'{width//gcd}:{height//gcd}'
+                                            size_info = f'Image Size: {width} x {height} , Ratio: {ratio} , {lcm_ratio}'
+
+                                            recommended_sizes = [
+                                                '704*1408', '704*1344', '768*1344', '768*1280', '832*1216', '832*1152',
+                                                '896*1152', '896*1088', '960*1088', '960*1024', '1024*1024', '1024*960',
+                                                '1088*960', '1088*896', '1152*896', '1152*832', '1216*832', '1280*768',
+                                                '1344*768', '1344*704', '1408*704', '1472*704', '1536*640', '1600*640',
+                                                '1664*576', '1728*576'
+                                            ]
```
mashb1t, line 245: this could be modules.config.default_aspect_ratio or do you explicitly only want to recommend core SDXL resolutions?

mashb1t (changes requested):
@xhoxye thanks for contributing, will check this out in detail in a few hours. Please find one commend attached.

mashb1t:
@xhoxye I've applied some adjustments, please check again

mashb1t (approved):

xhoxye:
Thank you very much for your help, although it is a little different from what I had in mind at the beginning, I initially wanted to recommend the standard size of SDXL

mashb1t:
@xhoxye thanks for the feedback, that's what i assumed in https://github.com/lllyasviel/Fooocus/pull/2971#discussion_r1608493765 and what makes the most sense.
My reasoning was that people might train LoRAs with specific aspect ratios or bucketing with other resolutions, but when the base model isn't trained these aspect ratios aren't the best choice anyways.
Users can still decide to just choose another available aspect ratio.

xhoxye:
I modified the use of textbox to display information to facilitate the translation of titles
QQ截图20240524213242
QQ截图20240524213348
`                                desc_image_size = gr.Textbox(label='Image Size and Recommended Size', elem_id='desc_image_size', visible=False)`

mashb1t:
Will include it in a few hours. Please add a new PR to develop if you want to be the author, if not i'll just cherry pick from your branch.

PR #2985, opened 2024-05-23
Update docker.md to prevent  #2836 on windows

stefanseeger:
On windows git is converting the line ending LF per default to CRLF, which will break the docker compose see #2836
To enforce git clone with line ending lf I updated the docker.md


docker.md
```diff
@@ -13,7 +13,8 @@ The docker image is based on NVIDIA CUDA 12.4 and PyTorch 2.1, see [Dockerfile](
 
 ### Running with Docker Compose
 
-1. Clone this repository
+
```
stefanseeger, line 16:
```suggestion
```

stefanseeger:
Hi @mashb1t ,
I just found that it might be a better solution to add a ` .gitattributes` file enforcing the line endings for everyone using this repo.
See docs: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings
What do you think about it?

mashb1t:
@stefanseeger yes, sounds reasonable. Please go ahead and add the file, if not i can also do it.

mashb1t:
done in https://github.com/lllyasviel/Fooocus/pull/2991

PR #3206, opened 2024-06-29
RestAPI for Fooocus

mrhan1993:
Full Fooocus with Fastapi


mashb1t:
@mrhan1993 keep up the good work! 👍

mrhan1993:
> @mrhan1993 keep up the good work! 👍

Due to development experience and standardization issues, I am uncertain whether it should be merged into the main branch. Do you have any suggestions on this matter?

mashb1t:
Pro same repository
* one single source of truth
* unified docs
* joint community
* reusability of code / deduplicate code
* release includes both updates for Fooocus and REST API

Contra same repository:
* responsibility for both Websockets and REST API in one repository
* more conflicts in the code
* increased testing effort => increased time to release also for small improvements / fixes

I personally do not have the resources to handle both projects in parallel, so coordination would be key, which then itself can be another disadvantage due to loosing speed on the features, but also an advantage if you're joining Fooocus development in general.

I'm torn between the two options, leaning more towards keeping separate repositories and synching more if you'd be open for that. What do you think?

mrhan1993:
What you said is exactly what I'm worried about. Considering my work situation, the merge might bring you more workload. Although I'm also wavering between the two, I tend to exist as an independent repository. Maybe it's not always the latest version, but it's more free.

mashb1t:
Ok, then let's keep it as is. Wishing you all the best!

PR #3216, opened 2024-06-30
feat: add pony preset and vae downloads

mashb1t:
uses default SDXL VAE, but the model doesn't include it

DOES NOT SUPPORT INPAINT with any inpaint engine as inpaint models have been trained on SDXL. Works with inpaint mode Improve Detail (face, hand, eyes, etc.)` though, as this doesn't use any inpaint engine. (same as playground model, but with better compatibility as no special sampler is needed)

Examples (fantasy themed & unmodified, metadata in images):

| Tiefling Shaman | Mouse Wizard | Bear Druid |
|--------|--------|--------|
| image - 2024-06-30T160236 664 | image - 2024-06-30T163504 951 | |


mashb1t:
Part of v2.5.0

PR #3218, opened 2024-06-30
Add french translation

LaBicHetteI:
French translation


LaBicHetteI:
Thanks for the validation

Issue #3161, opened 2024-06-18
[Feature Request]: hash generation


kalle07:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits

### What would your feature do?

Is there an intention to "pimp" hash generation with multiprocessing/filesplitting?
or an option generate hash from all files

### Proposed workflow

...

mashb1t:
Hashes are generated only once per model and only once for each start of the application.
Currently there is no need, at least no issues are known. Do you encounter any difficulties?

kalle07:
no difficulties but need time ,so maybe it give some moves ;)

i mean only an option generate hash from all files at once (so i can go trink a cup of coffee)

mashb1t:
That's a relatively easy task in python, so if you are familiar feel free to add a lazy loading for a new method call for each of the models and LoRAs, calculate the hash and write it to a file. If it already exists, then load hashes from that file. This way we can prevent recalculating the hashes each time after Fooocus restart, so it could benefit everyone.
I'd estimate about 2 hours for the implementation.
Do you want to take over the implenentation of this feature?

kalle07:
oh no sry iam not a programmer   : D

but this option (if i have 20 models in the folder no one with hash) calc hash from all models, i saw somewhere, ithnk that can be easy implemented....

mashb1t:
@kalle07 done, check out https://github.com/lllyasviel/Fooocus/pull/3221.

mashb1t:
part of https://github.com/lllyasviel/Fooocus/commit/f2a21900c64c0bd4ed925420ea84e643f85d292b

PR #3221, opened 2024-07-01
feat: persistent model hashing

mashb1t:
closes https://github.com/lllyasviel/Fooocus/issues/3161

Adds a persistent hash cache file to save time on hash generation after restarting Fooocus.
Use `--rebuild-hash-cache` to generate missing model and LoRA hashes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ __pycache__
 *.partial
 *.onnx
 sorted_styles.json
+hash_cache.json
 /input
 /cache
 /language/default.json
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ __pycache__
 *.partial
 *.onnx
 sorted_styles.json
-hash_cache.json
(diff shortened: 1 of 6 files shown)


kalle07:
hey...

--rebuild-hash-cach
ok, maybe an option 4 or 8 same time so each core can handle one model or lora   ;)

mashb1t:
@kalle07 done, but keep in mind that most likely the drive is going to be the bottleneck now.

I have a Ryzen 9 5900X 12 core / 24 thread CPU and a Sabrent Rocket Gen4 (1TB) M.2 NVMe with 5000 MB/s read & write and the SSD is now bottlenecking.

kalle07:
ah ... you are right, but maybe 100 loras could be faster  ? ^^

mashb1t:
part of https://github.com/lllyasviel/Fooocus/commit/f2a21900c64c0bd4ed925420ea84e643f85d292b

PR #3238, opened 2024-07-04
Rename readme.md to Anime giro without clothes

Samuppppp:


mashb1t:
Speedrun getting banned any%?

Screenshot_20240704-205939

PR #3263, opened 2024-07-10
Make --always-offload-from-vram actually work properly, fixes #3257

### LINKED ISSUE #3257: [Bug]: ROCm Fooocus doesn't garbage collect allocated VRAM
### Checklist

- [ ] The issue has not been resolved by following the [troubleshooting guide](https://github.com/lllyasviel/Fooocus/blob/main/troubleshoot.md)
- [ ] The issue exists on a clean installation of Fooocus
- [ ] The issue exists in the current version of Fooocus
- [ ] The issue has not been reported before recently
- [ ] The issue has been reported before but has not been fixed yet

### What happened?

On ROCm / amdgpu, Fooocus doesn't garbage collect used VRAM even after several hours. This means that other applications, such as other AI image generators, cannot use the VRAM and give "out of memory" errors.

### Steps to reproduce the problem

1. With ROCm with an AMD GPU, use Fooocus as normal - generate a few random pictures with default settings.
2. Wait a few hours then run `radeontop`. See that VRAM allocation is still many GB.
3. Use some other AI tool such as InvokeAI - generate a few random pictures with default settings.
4. See that this other AI tool gives Out of VRAM errors.
5. Close Fooocus and repeat (3)
6. The other AI tool succeeds.

### What should have happened?

1. Fooocus should release VRAM after it is finished generating images.

### What browsers do you use to access Fooocus?

_No response_

### Where are you running Fooocus?

Locally

### What operating system are you using?

Debian GNU/Linux

### Console logs

```Shell
No relevant logs in Fooocus.

As described above, behaviour is observed empirically via other means, i.e.

1. `radeontop` VRAM usage before and after shutting down Fooocus
2. console logs of {other AI tool} before and after shutting down Fooocus.
   - before: OutOfMemoryError specifically VRAM
   - after: works fine
```

### Additional information

_No response_

infinity0:


mashb1t (changes requested):
this is in contrast with the comment in https://github.com/infinity0/Fooocus/blob/a3880034e476a4fb5c1c3ba17635b08e5548fe88/ldm_patched/modules/model_management.py#L769, please confirm this makes the handling better instead of worse.

infinity0:
Need @lllyasviel to comment as I have no idea what the original comment is referring to regarding "worse". Works fine over here.

PR #3318, opened 2024-07-19
[Feature Request]: Added Clear Button that clears the prompt textbox.

Glought:
A Simple Button that clears the prompt textbox simplify the process when copying and pasting prompts.
foooocus_clear_button

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -99,7 +99,7 @@ div:has(> #positive_prompt) {
 }
 
 .advanced_check_row {
-  width: 330px !important;
+  width: 360px !important;
 }
 
 .min_check {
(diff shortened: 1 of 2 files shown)


mashb1t:
This is a won't do as the button is not in line with the design philosophy as well as it doesn't offer much benefit as one can just delete the prompt. Feel free to keep it in your fork though and optimize by directly returning empty steing without variable allocation.

Glought:
Honestly what is the design philosophy I can't find any info on it.

mashb1t:
It's not a document but Fooocus aims to be the open source alternative to Midjourney, and they don't have a clean prompt button as this simply is your input field in discord. Also the button doesn't integrate seamlessly in the UI as it's not next to already existing buttons but in a line of checkbox elements.

Glought:
ok that makes sense. I placed the button there because I didn't want to squish the prompt textbox. I'll just keep my changes to my fork.

PR #3427, opened 2024-08-02
Use weights_only for loading

kit1980:
`torch.load` without `weights_only` parameter is unsafe. Explicitly set `weights_only` to False only if you trust the data you load and full pickle functionality is needed, otherwise set `weights_only=True`.

If `weights_only=True` doesn't work for some cases, then explicit `weights_only=False` should be used.

Found with https://github.com/pytorch-labs/torchfix/

### DIFF
--- a/extras/BLIP/models/blip.py
+++ b/extras/BLIP/models/blip.py
@@ -216,9 +216,9 @@ def is_url(url_or_filename):
 def load_checkpoint(model,url_or_filename):
     if is_url(url_or_filename):
         cached_file = download_cached_file(url_or_filename, check_hash=False, progress=True)
-        checkpoint = torch.load(cached_file, map_location='cpu') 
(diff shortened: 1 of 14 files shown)


mashb1t:
Thanks for your contribution and the hint.

Relates to
- https://github.com/google/gemma_pytorch/issues/1
- https://github.com/DLR-RM/stable-baselines3/issues/1852

From https://pytorch.org/docs/stable/generated/torch.load.html:
> Warning
[torch.load()](https://pytorch.org/docs/stable/generated/torch.load.html#torch.load) unless weights_only parameter is set to True, uses pickle module implicitly, which is known to be insecure. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling. Never load data that could have come from an untrusted source in an unsafe mode, or that could have been tampered with. Only load data you trust.

mashb1t (approved):

Issue #3192, opened 2024-06-25
[Feature Request]: Support for DORA and other types of LORAs


santiagosfx:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits

### What would your feature do?

Compatibility with DORAs and other types of LORAs to generate images

### Proposed workflow

Work as usual

mashb1t:
Please provide more details about what your expectation is with this feature and what does / does not work. The description can e.g. be enriched with references to papers or models.

santiagosfx:
According to the tests I have done, Foocus simply does not detect the Doras. I tried with a concept, and a drawing style, and it failed to be detected. It generated images, but it was as if the dora was not there. Even if I had loaded it and used the trigger word. 
As far as I know, Dora is basically a potential successor to Lora, or a variant. I'll leave more information about it in the commentary, and some Dora models from Civitai as examples.
https://www.reddit.com/r/StableDiffusion/comments/1asxk3u/dora_weightdecomposed_lowrank_adaptation/ https://sebastianraschka.com/blog/2024/lora-dora.html

Sample civitai lora model: https://civitai.com/models/483185/takagi-san-karakai-jouzu-no-takagi-san

mashb1t:
Thanks for elaborating. DoRA support is also included in https://github.com/lllyasviel/Fooocus/pull/3084, but sadly i need some extra time and/or support here. Any help is welcome!

raochinmay6:
sorry for the comment here but can you add an image to video generation feature in fooocus like cumfyui , it would be so nice, I personally like fooocus so much because of its easy webui interface, but comfyui is a little bit complex for me ... can it will be happen in future ?

mashb1t:
@raochinmay6 this is off-topic, but Fooocus does not and will not have video generation capabilities as it's an image AI tool only. There are already tools out there who specialize in video generation.

PR #3454, opened 2024-08-05
Test update ldm patched

IPv6:
managed to cleanup errors and now it is working with updated ldm_patched

calculate_sigmas - some type unification added
some parameters (default_image_only_indicator) now defaults to None to avoid exceptions
some deprecated method unwrapped
samples/schedulers - tested, seems to be ok
- karas, euler, heun, restart, some others && samplers - no problems
- uni_pc/uni_pc_bh2: updated from latest comfy, original code had function parameters mismatch
controlNets - tested, seems to be ok

did not tested: Refiners (SDXL/SD15), Inpainting


mashb1t:
@IPv6 thanks a lot, much appreciated! There still aeem to be unresolved MRs, which you can solve by merging main to your branch and to resolve them locally.

IPv6:
@mashb1t  Updated with conflict resolve. Uncertain about some changes:

- removed opencv-contrib-python==4.8.0.74 
due opencv-contrib-python-headless==4.10.0.84 already present

mashb1t:
Thanks for the updates, can check this on the weekend!

mashb1t:
LGTM after fixing a bug with favicon (missing args). It's quite the spam in the logs due to logging all requests done by HTTPX, but manageable.

2 concerns:
* images before and after are not 100% the same, but 99% (like using xformers). This might have to do with internal changes of the code.
* It was a mess to do package updates as some users, mostly non-tech-people (Fooocus is the "simple" interface), are not able to do updates themselves without massive support despite providing [update notes on how to update](https://github.com/lllyasviel/Fooocus/releases/tag/v2.5.0).

Sadly i don't have the time, neither can i implement new features nor can i support as extensively as last time with the 2.5.0 enhance upgrade until November, as there's a lot going on in my private and work life. During this period, nobody will be able to do code updates as i'm sadly the only maintainer and i can only assume lllyasviel has ditched the project despite mentioning to come back after the release of SD3 (see https://github.com/lllyasviel/Fooocus/discussions/2154#discussioncomment-8389770).
=> Merging this feature now could result in reports of unknown errors, which nobody would be able to fix.

I'm open to your suggestions, hope you understand the situation.

Thank you once more for your support, much appreciated!!

mashb1t:
@IPv6 fyi I've decided to step back as maintainer of Fooocus, details see https://github.com/lllyasviel/Fooocus/discussions/3504.
In a nutshell: no time, doing other things, as already mentioned above.

IPv6:
Totally understandable, Fooocus "simplicity aura" has this drawback of endless flow of end-user problems // And such update can mess a lot, most probably in some hidden cases.

Anyway, I think any fork that wish to implement something ldm-fresh will be able to use this patch as a starting point. For me Fooocus already quite mature and frankly there is nothing really needs to be added given perfect Fooocus focus. It already works :)

mohamed-benali:
I got a question that I would appreciate if you could answer.

If I switch to this branch and run Fooocus it is going to run? Is so, could DoRa be easily added? Or that would require a lot of work?

Thanks in advance!

mashb1t:
@mohamed-benali this branch supports DoRA out of the box, you can switch to it and just select it as any other LoRA.

mohamed-benali:
Thats amazing!

Since there will be very few updates, unless @lllyasviel takes this back *wink wink*, I will start using this branch. Hopefully I won't find any blocking bug.

Thanks for your work!! It's very appreciated!

mohamed-benali:
After pocking around and trying this version I have to say I am very impressed!

It's sooo much faster, specially the initial setup which is almost 0 now. And the actual generations are faster too.

Thank you for updating this @IPv6 @mashb1t , it's very appreciated!!

IPv6:
> (only issue seems to be the anime inpaiting after step 24ish of 30 steps, which seems to error. But I can use the official version to inpaint and this one to generate faster)

i also hit this some days ago - and fixed it, the problem was in synthetic refiner for inpainting... this inpainting stuff blows the mind. Added a fix to this branch - https://github.com/IPv6/Fooocus4BL/commit/03fabf0b4813f28c7299a707689b6a802acf1a1b

Theoretically you can apply it locally, if needed. with it inpainting works (and presumably SDLX refiners in general)

mohamed-benali:
Thank you so much @IPv6. It's working for me too :)


Just have 1 question. Are DORAs supposed to work the same as LORAs? They don't seem to work for me, so just asking to know if I am missing something or they just don't work. (thanks in advance!)

IPv6:
Hard to tell, with new ldm they should work (as mashb1t noted), but ldm is just the basement of all that machinery. There are also higher-level checks, and i suspect they need some upgrading to recognise DORAs as valid LORAs models.

There is no such specific fixes in this patch, unfortunately. But it should be easy to add, since ldm do all the heavy lifting.

mohamed-benali:
Do you know which module/files need to be touched? If it's that easy I might try to see if I can do it.

IPv6:
Hard to tell, not really using them // Looking at code there is suspicious lora parsing in "match_lora" method in core.py. Possibly DORA have some unexpected internal layers that are not recognized, this is not a part of ldm_patched so need to be updated by someone who familiar with torch model insides //

mohamed-benali:
Thanks for the insight! Just as a note, the DORA never errors when I select it or generate an image. It's just that it has no effect.

mohamed-benali:
I have made some testing and it appears that DORAs give a `model mismatch` on here: `if len(lora_unmatch) > 12:`

After investigating more I found that it's missing some keys on `def match_lora(...)`. 

I added the DORAs portion from here (all the code across the `def load_lora(...)`):

https://github.com/comfyanonymous/ComfyUI/blob/f067ad15d139d6e07e44801759f7ccdd9985c636/comfy/lora.py#L46

I added it locally and it worked for me!! DORAs work properly now!!

I would push it to this branch but I cannot since I don't have edit permissions.

IPv6:
> I added it locally and it worked for me!! DORAs work properly now!!
> I would push it to this branch but I cannot since I don't have edit permissions.

Whoah, pretty cool! So it was lora parsing after all
Congrats!!!

If you attach your updated lora.py - i can push it to the branch, just to keep all updates in one place

IPv6:
> Here it is:

Thanks! Pushed it to the branch - https://github.com/IPv6/Fooocus4BL/commit/1157a1df68e005568c0f449e278042833cea6485

mohamed-benali:
Perfect!! With the fixes for Inpainting and DORAs this will be my main source for Fooocus. No need to switch between main and this one. Good job @IPv6!! 

@santiagosfx I saw in here https://github.com/lllyasviel/Fooocus/issues/3192 you were asking for DORAs support. If you use this branch they will work correctly!

santiagosfx:
> Perfect!! With the fixes for Inpainting and DORAs this will be my main source for Fooocus. No need to switch between main and this one. Good job @IPv6!!
> 
> @santiagosfx I saw in here #3192 you were asking for DORAs support. If you use this branch they will work correctly!

Sorry for the newbie question, but how do I use a branch instead of the main repository?
Specifically in Google Collab.

IPv6:
> Sorry for the newbie question, but how do I use a branch instead of the main repository? Specifically in Google Collab.

Don`t know regarding colab, but for using branch you need to checkout project from top commit of the branch - https://github.com/IPv6/Fooocus4BL/commit/1157a1df68e005568c0f449e278042833cea6485
You need to be familiar with git

santiagosfx:
> > Sorry for the newbie question, but how do I use a branch instead of the main repository? Specifically in Google Collab.
> 
> Don`t know regarding colab, but for using branch you need to checkout project from top commit of the branch - [IPv6@1157a1d](https://github.com/IPv6/Fooocus4BL/commit/1157a1df68e005568c0f449e278042833cea6485) You need to be familiar with git

I managed to get the branch running, but only halfway.
It seems to be missing a parameter to make it work on Colab, called "--always-high-vram".
What it does is to better manage memory. Without this parameter, I end up running out of memory before I finish generating the image.

IPv6:
>  called "--always-high-vram". What it does is to better manage memory

This param was changed in Comfy, so updated ldm require new setting instead of old one. try --highvram --use-pytorch-cross-attention --force-fp16

santiagosfx:
> > called "--always-high-vram". What it does is to better manage memory
> 
> This param was changed in Comfy, so updated ldm require new setting instead of old one. try --highvram --use-pytorch-cross-attention --force-fp16

It keeps giving me the same error, the only difference now is that it does it in the middle of loading the models.

kiruthikpurpose (approved):

mohamed-benali:
Any idea how hard it would be to add support for SD3.5? Would be as difficult as adding the DORA support? Or far higher?

IPv6:
> Any idea how hard it would be to add support for SD3.5? Would be as difficult as adding the DORA support? Or far higher?

SD3.5 has different architecture comparing to SDXL, so it is the same as with Flux - almost full rewrite is needed //

PR #3759, opened 2024-11-28
full path to load safety checker model cause OSError

mrhan1993:
OSError: Incorrect path_or_model_id: 'D:\ComfyUI\models\safety_checker\stable-diffusion-safety-checker.bin'. Please provide either the path to a local folder or the repo_id of a model on the Hub.

it should be a directory and model named pytorch_model.bin


mashb1t:
No, the filename `stable-diffusion-safety-checker.bin` is not used by accident. All of the config options of that group are directories, but i understand what your intention is. Nevertheless the code is implemented as intended.

PR #3841, opened 2025-01-16
Fix: Readme path

M123-dev:
Quick for a path in the Readme.

The correct File is in the `Fooocus/` folder as correctly mentioned a couple of lines above

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -215,7 +215,7 @@ Then run the `run.bat`.
 
 AMD is not intensively tested, however. The AMD support is in beta.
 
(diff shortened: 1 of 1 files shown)


mashb1t:
Thanks for the adjustments. I assume your adjustments do mix up backwards and forwards slash and are inconsistent in the command. Yould you double-check and use `\` instead?

M123-dev:
@mashb1t thanks for your review, yeah you are right `\` fits better, fixed.

PR #4124, opened 2025-11-26
Optimisation for RTX 5060TI

liubomyr-lototskyi:
I noticed a rapidly growing trend where people are actively switching to RTX 5090, RTX 5070, 5080, etc., and suddenly realise that their familiar software does not work due to CUDA incompatibility.
This applies to stable-diffusion-webui, fooocus, and a number of other popular programmes. What to do in this situation? Install CUDA 12.8, install library versions that support it, such as torch nightly, and compile packages yourself if they do not yet have native support for the Broadwell generation (RTX 50xx).

After spending more than 20 hours testing different versions of CUDA, torch, torchvision, and various versions of wheels, I finally found a working setup.


launch.py
```diff
@@ -29,7 +29,7 @@
 def prepare_environment():
     torch_index_url = os.environ.get('TORCH_INDEX_URL', "https://download.pytorch.org/whl/cu121")
     torch_command = os.environ.get('TORCH_COMMAND',
-                                   f"pip install torch==2.1.0 torchvision==0.16.0 --extra-index-url {torch_index_url}")
+                                   f"pip install torch==2.7.1+cu128 orchvision==0.22.1+cu128 --extra-index-url {torch_index_url}")
     requirements_file = os.environ.get('REQS_FILE', "requirements_versions.txt")
```
lcretan, line 33: I understood your purposes and efforts.

Then, the stable PyTorch is now version 2.9.1.
Also, the stable CUDA 12.x is now 12.9.
They are fixed many bugs and enhanced to adapt your purposes.

Furthermore, cpython 3.10 has Intel distributed optimization libraries now provided as the pip packages, not Intel's original python fork.

So, to realize your purposes effectively and efficiently, I propose the combination: cpython 3.10 + Intel libraries + torch 2.9.1 + CUDA 12.9 although their wheels are often hidden, but we can setup them by the each install option.

If you have already tested them and found degradings , I need to change my decision.
Because I'm now using and testing them with RTX 4070 to enhance the original Fooocus, not fork original  one, even slightly, to contribute the original Fooocus team who have built and maintained this state-of-the-art and out-of-box great package.

Regarding CUDA 13.0, the structure is drastically changed and still has many bugs although PyTorch already supports CUDA 13.0.

Thanks,

tommy3210:
Dang, thanks man.

